PR #3, opened 2026-07-02
fix(extension): never return a borrowed tab into another session's Agent Window

hobostay:
## Problem

When a borrowed tab's original window has been closed, `tool.tab_return` (and `tool.session_stop`'s auto-return path) needs a fallback window to put the tab back into. The picker in `chooseFallbackWindow` used:

```ts
const last = await windowsApi.getLastFocused({ windowTypes: ["normal"] });
if (lastId !== null && lastId !== ctx.agentWindowId) {
  return { windowId: lastId, index: -1 };
}
```

It only excluded **this** session's Agent Window. But Agent Windows are created with `type: "normal"` (`session-manager/agent-window.ts`), so they survive the `windowTypes: ["normal"]` filter. If another session's Agent Window happened to be last-focused (common — agent windows grab focus), a returned user tab could be moved into **another session's Agent Window**.

Consequences:
- **Cross-session sandbox break:** the receiving session now sees the tab as a native Agent-Window tab, so its write tools (`click`/`fill`/`evaluate`/…) pass `enforceAgentWindow` and can act on a tab the user never offered to that session.
- **Silent user-data loss:** when that session later stops, `SessionManager.stop` closes its Agent Window, destroying the user's tab along with it. The owning session's `borrowedTabs` no longer references it, so it can't be returned.

This contradicts the project's stated posture ("leave the rest of your browser alone") and the cross-session isolation rule enforced everywhere else (e.g. `handleTabList`, `resolveTargetTab`, `listConfirmationCandidates`).

## Fix

Thread an `isAgentWindowId: (windowId: number) => boolean` predicate through `returnBorrowedTab` → `chooseFallbackWindow`, and skip **every** live Agent Window as a fallback target (not just the current session's). This mirrors the existing filter in `borrow-confirmation.ts`'s `listConfirmationCandidates`.

- `TabManagementDeps` gains an optional `isAgentWindowId`.
- The production handlers (`handleTabReturn`, `handleSessionStop`) default it to `(windowId) => manager.findByWindowId(windowId) !== null`, so the fix is active by default with no dispatcher wiring change.
- Direct/test callers keep the legacy `() => false` default.

## Tests

Two regression tests in `tabs.test.ts` (both **fail** on the unpatched code, verified by reverting the source change):

1. `does not fall back into another session's Agent Window` — two sessions (windows 100 & 200); session A returns a borrowed tab whose origin window is gone while `getLastFocused` returns session B's window 200. Asserts a fresh window is created (777) and the tab is never moved to 200.
2. `returnBorrowedTab honours an explicit isAgentWindowId predicate` — direct unit coverage of the predicate plumbing.

Full suite: `361 passed (34 files)`, no regressions. Biome/tsc clean for the touched files (a few pre-existing format/type findings in unrelated files are untouched).

## Scope

Deliberately minimal and focused on this one clearly-correct bug. I audited the broader sandbox/session/lifecycle surface and have a shortlist of other findings (e.g. borrow-confirmation fail-open defaults, the borrowed-tab-destroyed-on-window-close path, daemon WS event ownership) that I'm happy to follow up on in separate PRs if the maintainers are interested.

### DIFF
--- a/apps/extension/src/tools/__tests__/tabs.test.ts
+++ b/apps/extension/src/tools/__tests__/tabs.test.ts
@@ -613,6 +613,65 @@ describe("handleTabReturn", () => {
     expect(winSpies.create).toHaveBeenCalledOnce();
   });
 
+  it("does not fall back into another session's Agent Window (creates a new window instead)", async () => {
+    // Regression: a returned borrowed tab whose original window is gone
+    // must never be parked in *another* session's Agent Window. Agent
+    // Windows are `type: "normal"`, so getLastFocused({windowTypes:
+    // ["normal"]}) can return session bb22's window (200). Without the
+    // isAgentWindowId guard the tab would be moved into window 200,
+    // letting bb22 write to it and destroying it when bb22 stops.
+    const sm = new SessionManager({ agentWindow: fakeAgentWindow([100, 200]) });
+    const ctx = await sm.start("aa11");
+    await sm.start("bb22"); // owns Agent Window 200
+    ctx.borrowedTabs.set(7, { tabId: 7, originalWindowId: 300, originalIndex: 4 });
+    const state: FakeTabState = {
+      tabs: new Map([[7, { id: 7, windowId: 100 } as chrome.tabs.Tab]]),
+      nextTabId: 50,
+      windowsClosed: new Set([300]),
+    };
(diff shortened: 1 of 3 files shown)


BB-fat:
Thanks for the great catch and clean fix, @hobostay !

PR #4, opened 2026-07-02
fix(daemon): reject cross-browser session events from non-owners

hobostay:
## Problem

The daemon dispatched inbound `session.window_closed` and `session.user_interrupt` events using **only the event payload** — the sending browser's identity was never checked against the session's owner (`crates/bsk-cli/src/daemon/ws.rs`, `handle_inbound_text` → `handle_session_window_closed` / `handle_session_user_interrupt`).

Because `handle_inbound_text` already had the sending `client: &Arc<BrowserClient>` in scope but did not pass it down, neither handler could verify ownership. Result: any connected extension could send an event naming **another browser's** `session_id` and the daemon would act on it:

- `session.window_closed` → `forget_session` tears down the victim's session (drops it from the registry, removes its tool queue + interrupt marker).
- `session.user_interrupt` → `cancel_session` cancels every inflight tool of the victim's session and arms its interrupt marker.

This is a cross-tenant integrity / DoS hole that is exploitable **even between two legitimate extensions** (e.g. two browsers paired to one daemon), independent of the separately-tracked WS Origin allow-list gap. With the Origin gap it extends to any same-machine process that completed the WS handshake.

## Fix

Thread the sender's `BrowserId` (`client.id`, already in scope) into both handlers and ignore the event when the named session is live and owned by a different browser:

```rust
if let Some(session) = state.sessions.get(&session_id)
    && session.browser_id != *sender
{
    warn!(...);
    return;
}
```

A session that is already gone is left to the existing no-op `forget` path (no behaviour change for the legitimate disconnect race).

## Tests

Four new ownership tests in `ws.rs` (`session_user_interrupt_tests`): owner-honoured and non-owner-ignored for **each** event. The two pre-existing interrupt tests were updated for the new signature (they register no live session, so the guard's "unknown → allow" path preserves their original behaviour).

Verified: full `cargo test -p bsk` green — 165 lib tests + all integration suites (including `session_user_interrupt`, `cancel_forwarding`, `per_session_queue`, `ws_handshake`), 0 failures. No new clippy warnings (the two existing ones are in untouched code).

## Scope

Minimal and focused. This is part of a shortlist of findings from a broader sandbox/session/lifecycle audit; separate PRs handle the extension-side tab-return cross-session bug and (pending) the borrow-confirmation fail-open edge cases.

### DIFF
--- a/crates/bsk-cli/src/daemon/ws.rs
+++ b/crates/bsk-cli/src/daemon/ws.rs
@@ -416,10 +416,10 @@ async fn handle_inbound_text(state: &Arc<DaemonState>, client: &Arc<BrowserClien
         }
         Frame::Event(ev) => match ev.event {
             bsk_protocol::EventKind::SessionWindowClosed => {
-                handle_session_window_closed(state, &ev.payload);
+                handle_session_window_closed(state, &client.id, &ev.payload);
             }
             bsk_protocol::EventKind::SessionUserInterrupt => {
-                handle_session_user_interrupt(state, &ev.payload);
+                handle_session_user_interrupt(state, &client.id, &ev.payload);
             }
             other => {
                 debug!(event = ?other, "event received (no handler yet)");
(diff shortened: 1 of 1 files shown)


BB-fat:
LGTM

PR #5, opened 2026-07-02
fix(extension): don't fail-open the borrow gate on a transient tabs.get error

hobostay:
See file diff. Fixes an undocumented fail-open in requestBorrowConfirmation: chrome.tabs.get(tabId) throwing returned true (approve) instead of falling back to a generic title and still showing the confirmation overlay/notification. The tab was already fetched earlier by the borrow pipeline, so this is a transient failure being silently auto-approved with no UI. Adds two regression tests (fail on unpatched code). Full suite green (361 tests). Sibling PRs #3 and #4.


BB-fat:
LGTM

PR #15, opened 2026-07-10
fix(extension): require borrowing before tab access

NianJiuZst:
## What problem this solves

BrowserSkill promises that an existing user tab must be explicitly borrowed before the agent touches it. The write-oriented handlers already enforce that boundary, but the shared explicit `tab_id` resolver treated every normal user window as visible to the active session.

That meant an agent could list user tabs, take a tab id, and call observation tools such as `snapshot`, `get-html`, `screenshot`, or `console` without showing the borrow confirmation or moving the tab into the Agent Window. The same resolver is also used by evaluate, wait, navigation, human-loop, and interaction handlers, so the inconsistency existed at the common authorization boundary rather than in one isolated command.

This was especially risky for already-authenticated pages because a read-only command could expose page content before the user had granted per-tab access. Existing tests encoded the old behavior by expecting screenshot and console access to succeed against a tab in a regular user window.

## How this fixes it

The shared target resolver now requires every explicit tab to belong to the current session's Agent Window. A pre-existing user tab becomes eligible only after `tab_borrow` moves it into that window. Tabs owned by another session continue to return `not_found`, so this change does not weaken cross-session isolation or reveal another session's window membership.

The resolver returns the existing structured `permission_denied` / `agent_window_scope` error for an ordinary user tab. This gives callers a consistent, actionable explanation: borrow the tab before reading or interacting with it.

Tests now cover both sides of the boundary:

- an explicit user tab is rejected before CDP or screenshot capture is invoked;
- an explicit tab in the current Agent Window remains accessible;
- screenshot and console regressions assert that no browser-side read occurs on denial.

## User impact

Agents can still discover user tabs through `tab list` so they can identify a borrow target, but they cannot inspect page content until the user-approved borrow flow completes. Agent-created tabs and already-borrowed tabs continue to work because both live in the session's Agent Window.

This aligns runtime behavior with the README and privacy-policy promise that pre-existing tabs stay outside agent control until explicitly borrowed.

## Validation

- `corepack pnpm --filter @browser-skill/extension test -- src/tools/__tests__/shared.test.ts src/tools/__tests__/observation.test.ts src/tools/__tests__/console.test.ts`
  - 35 test files passed
  - 375 tests passed
- `git diff --check`

## CI Baseline Compatibility

## CI Reliability Follow-up

### DIFF
--- a/apps/extension/src/tools/__tests__/console.test.ts
+++ b/apps/extension/src/tools/__tests__/console.test.ts
@@ -72,14 +72,14 @@ describe("handleConsole", () => {
     expect(deps.consoleEntriesSince).toHaveBeenCalledWith(7, undefined, 50, 1000, false);
   });
 
-  it("reads an explicit tab and forwards bounded options", async () => {
+  it("rejects an explicit user tab before touching CDP", async () => {
     const sm = new SessionManager({ agentWindow: fakeAgentWindow([100]) });
     await sm.start("aa11");
     const deps = makeDeps({
       get: vi.fn(async () => ({ id: 9, windowId: 200, active: true }) as chrome.tabs.Tab),
     });
 
-    await handleConsole(
+    const res = await handleConsole(
       sm,
       {
         session_id: "aa11",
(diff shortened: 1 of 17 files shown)


BB-fat:
Hi @NianJiuZst — thanks for the PR and the careful thinking about the borrow model.

After review, we're closing this as **working as intended**: allowing read operations without borrowing is a deliberate product design decision. Read-only access (observation, screenshots, etc.) is intentionally non-exclusive so users and tooling can inspect tabs freely; the borrow/confirmation flow is reserved for actions that mutate state. Tightening reads to require borrow would break that intended balance.

Appreciate the contribution either way!

感谢贡献！经过评审：读操作允许不 borrow 是有意为之的产品设计——只读访问（观察、截图等）被刻意设计为不需要独占，borrow/确认流程只留给会改变状态的操作。因此本 PR 按「设计如此」关闭。

PR #17, opened 2026-07-10
fix(runtime): clean up sessions on disconnect

NianJiuZst:
## What problem this solves

BrowserSkill keeps a session registry in two processes: the daemon tracks the CLI-facing session id and dispatch queue, while the extension owns the Agent Window, ref store, CDP attachments, and borrowed-tab return metadata.

Before this change, losing the extension WebSocket caused the daemon to purge its side immediately, but the extension only cleared its handshake state. The local `SessionManager` stayed alive. Disabling the connection from the popup had the same behavior. After reconnecting, the CLI no longer knew those session ids, so it could not stop their Agent Windows or return borrowed tabs. Closing an orphan Agent Window could then close a user tab that was still inside it.

There was a second race in the daemon's reconnect generation guard. If a new socket registered before the old socket's cleanup completed, the guard correctly preserved the new browser registration, but it also preserved sessions that the extension no longer had. That produced the inverse split: daemon rows with no matching Agent Window.

## How this fixes it

This change adopts an explicit safety policy: transport loss terminates local browser sessions instead of attempting to preserve unverifiable in-memory state across the disconnect.

The extension now uses the same teardown path as a normal `session stop`:

1. return every borrowed tab to its original or fallback user window;
2. keep the Agent Window open if any return fails, avoiding user-tab loss;
3. clear the session ref store;
4. detach session-owned CDP targets;
5. close the Agent Window and remove the local SessionContext.

Cleanup is invoked in two lifecycle paths:

- before an intentional popup disconnect, so teardown completes before the socket closes;
- after an unexpected transport loss, before reconnecting.

The disconnect cleanup is coalesced, so an explicit disconnect event and the transport state callback cannot run two overlapping teardown passes. For unexpected loss, the controller first cancels WSTransport's pending automatic reconnect, performs local cleanup, and only then reconnects.

On the daemon side, a fresh handshake for an existing browser instance now purges any stale sessions and their queues/interrupt markers before replacing the browser registration. The generation guard still protects the new browser connection when the old socket task exits, but stale sessions are no longer carried across generations.

## User impact

After disabling BrowserSkill or losing the loopback connection, users no longer have hidden sessions that the CLI cannot manage. Borrowed tabs are returned before Agent Windows close. A reconnect starts from a clean state and the next automation task creates a fresh session id.

This intentionally trades transparent session continuation for a deterministic safety boundary. BrowserSkill session state is currently in-memory and cannot be proven consistent after a disconnect; preserving it was the source of the orphan behavior.

If Chrome refuses to return a borrowed tab, cleanup reports the failure and leaves that Agent Window open instead of risking data loss. The failure is logged for the user to resolve manually.

## Validation

  - Rust unit, integration, protocol, handshake, reconnect, and session tests passed
  - 36 test files passed
  - 377 tests passed
- `git diff --check`

New regression tests verify safe cleanup ordering, cleanup of multiple sessions, coalescing of overlapping disconnect notifications, intentional-disconnect ordering, unexpected-disconnect cleanup, and daemon-side stale-session removal during a same-instance reconnect.

## CI Baseline Compatibility

## CI Reliability Follow-up

### DIFF
--- a/apps/extension/src/entrypoints/background.ts
+++ b/apps/extension/src/entrypoints/background.ts
@@ -16,6 +16,7 @@ import {
 } from "@/lib/overlay-bridge";
 import { POPUP_PORT_NAME, type PopupInbound, type PopupOutbound } from "@/lib/popup-bridge";
 import { attachSessionsLiveFlag } from "@/lib/sessions-live-flag";
+import { createDisconnectCleanup } from "@/session-manager/disconnect-cleanup";
 import { attachSessionEventHandler } from "@/session-manager/event-handler";
 import { SessionManager } from "@/session-manager/manager";
 import {
@@ -41,6 +42,13 @@ export default defineBackground(() => {
   // stale `true` does not keep waking us on every page load until the
   // first mutation (review M4/M5 round 3 m-R3-1).
   void sessionsLive.refresh();
+  const cleanupAfterDisconnect = createDisconnectCleanup({
+    manager: sessions,
+    sessionStopDeps: { cdp },
+    onSessionsChanged: () => {
+      void sessionsLive.syncFromManager();
+    },
+  });
   const dispatcher = new ToolDispatcher({
     transport,
     sessions,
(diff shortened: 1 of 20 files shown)


BB-fat:
Thanks for the fix. Before merging, please address these lifecycle races:

- Disconnect cleanup should stop/cancel and await in-flight RPCs before snapshotting sessions. Otherwise `tab_borrow` or `session_start` can finish after cleanup, potentially closing a borrowed user tab or leaving an orphan Agent Window.
- Treat terminal handshake failures (for example, version incompatibility) differently from transient disconnects so they do not enter an immediate reconnect loop.
- On generation rollover, drain or cancel in-flight `session.start` reservations and compensate any window created by the old generation before purging state.

Please add race tests with delayed tab moves/window creation and a second handshake during an in-flight session start.

BB-fat:
Thank you @NianJiuZst — this is a well-thought-out fix: reusing the safe session-stop path for disconnect cleanup (returning borrowed tabs first, keeping the window when a return fails so user tabs are never lost), and the daemon-side handshake purge closes the reverse split-brain nicely. The tests are honest, which we appreciate. Merging. 🙏

感谢贡献！断连清理复用安全 stop 路径的设计很干净，合并了。

PR #18, opened 2026-07-10
fix(extension): make tool cancellation cooperative

NianJiuZst:
## What problem this solves

The extension previously implemented cancellation by racing every tool handler against a promise that rejected as soon as an AbortSignal fired. That made the RPC appear cancelled immediately, but it did not stop the handler that lost the race.

For read-only work this wasted CDP work and could corrupt later state: a cancelled snapshot could finish after a newer snapshot and replace the session RefStore with stale element ids. For mutating work the impact was more visible. Cancelling `tool.session_start` while `chrome.windows.create` was pending caused the daemon to discard its reservation, but the unresolved handler later created and registered an Agent Window that the daemon did not know about.

Several tab handlers already contained abort checks and compensation logic, but the dispatcher did not pass their AbortSignal. The result was a cancellation protocol that acknowledged cancellation without establishing that the underlying browser operation had stopped or rolled back.

## How this fixes it

The dispatcher no longer uses `Promise.race` to manufacture an early cancelled response. It still acknowledges the separate `cancel` request immediately, but the original tool RPC now waits for its handler to observe the signal and finish any necessary compensation before sending its final response.

Cancellation is wired through the previously disconnected paths:

- `session_start` receives the AbortSignal;
- tab create, close, select, borrow, and return receive the AbortSignal;
- screenshot, console, get-html, and snapshot receive cancellation checks around their awaited browser/CDP work;
- navigation, interaction, evaluate, wait, and human-loop handlers retain their existing signal plumbing.

`SessionManager.start` now treats Agent Window creation as a transactional operation. If cancellation arrives after `windows.create`, or if active-tab initialization fails, it removes the incomplete window and does not register the session. The session handler maps this AbortError to the structured `cancelled` result.

Snapshot checks the signal before replacing RefStore contents. A cancelled older snapshot can therefore finish its underlying CDP request without overwriting refs produced by later work.

## Cancellation boundary

Some Chrome APIs are irreversible after dispatch, such as removing a tab that Chrome has already accepted. The fix does not pretend those operations can be rolled back. Instead, it guarantees that BrowserSkill does not send the original RPC response while a handler is still running invisibly. Handlers with a reversible side effect, such as creating a tab or Agent Window, perform compensation before returning cancelled.

## User impact

Cancelling a slow command no longer leaves hidden Agent Windows or silently mutates the element-ref state after the CLI has moved on. The cancel control remains responsive because its own acknowledgement is immediate, while the cancelled tool's final lifecycle is now truthful and deterministic.

The change also fixes a non-cancellation leak: if Agent Window active-tab setup fails after window creation, the partial window is removed.

## Validation

  - 35 test files passed
  - 376 tests passed
- `corepack pnpm ext:build`
- `git diff --check`

Regression tests verify that:

- cancel acknowledgement is immediate but the original session-start RPC remains pending until cleanup finishes;
- the created Agent Window is removed and no SessionContext is registered;
- startup setup failures also remove incomplete windows;
- a snapshot cancelled during CDP capture leaves the previous RefStore untouched.

## CI Baseline Compatibility

## CI Reliability Follow-up

### DIFF
--- a/apps/extension/src/session-manager/__tests__/manager.test.ts
+++ b/apps/extension/src/session-manager/__tests__/manager.test.ts
@@ -57,6 +57,38 @@ describe("SessionManager", () => {
     await expect(sm.start("aa11")).rejects.toThrow(/already exists/);
   });
 
+  it("removes a newly created Agent Window when startup is aborted", async () => {
+    const aw = fakeAgentWindow();
+    let resolveCreate: (windowId: number) => void = () => {};
+    aw.createMock.mockImplementationOnce(
+      () =>
+        new Promise<number>((resolve) => {
+          resolveCreate = resolve;
+        }),
+    );
+    const sm = new SessionManager({ agentWindow: aw });
+    const controller = new AbortController();
+    const pending = sm.start("aa11", { signal: controller.signal });
+
+    controller.abort();
+    resolveCreate(777);
+
+    await expect(pending).rejects.toMatchObject({ name: "AbortError" });
+    expect(aw.removeMock).toHaveBeenCalledWith(777);
+    expect(sm.has("aa11")).toBe(false);
+  });
+
+  it("removes an incomplete Agent Window when active-tab setup fails", async () => {
+    const aw = fakeAgentWindow();
+    aw.ensureActiveTabMock.mockRejectedValueOnce(new Error("tab setup failed"));
+    const sm = new SessionManager({ agentWindow: aw });
+
+    await expect(sm.start("aa11")).rejects.toThrow("tab setup failed");
+
+    expect(aw.removeMock).toHaveBeenCalledWith(100);
+    expect(sm.has("aa11")).toBe(false);
+  });
+
   it("stop() closes the Agent Window and forgets the session", async () => {
     const aw = fakeAgentWindow();
     const sm = new SessionManager({ agentWindow: aw });
--- a/apps/extension/src/session-manager/manager.ts
+++ b/apps/extension/src/session-manager/manager.ts
@@ -25,6 +25,16 @@ export interface SessionManagerOptions {
   now?: () => number;
 }
 
+export interface SessionStartOptions {
+  signal?: AbortSignal;
+}
+
+function sessionStartAbortError(): Error {
+  const err = new Error("session_start aborted");
+  err.name = "AbortError";
+  return err;
+}
+
 /**
  * Owner of all live agent sessions inside the extension.
  *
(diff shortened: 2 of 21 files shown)


BB-fat:
Thanks for improving cancellation semantics. A few issues should be fixed before merging:

- The daemon currently releases the session queue as soon as cancellation is acknowledged, before the extension handler finishes compensation. Keep the forwarded request/session busy until the extension returns its final result or a cleanup timeout expires.
- Add `AbortSignal` checks after each awaited step and before every side effect in `tab_return` and multi-stage observation operations. If a fallback window was created, close it when cancellation wins.
- Do not silently swallow compensation failures; retain the orphan resource for retry or return an explicit cleanup failure.

Please add an end-to-end test with delayed compensation and ensure the full Rust test job completes successfully.

BB-fat:
Hi @NianJiuZst — thank you for this thoughtful piece of work. The problem it tackles is real and still present on main (orphaned shadow windows from cancelled `session_start`, and stale RefStore writes from cancelled snapshots), and the cooperative-cancellation design — honest responses after handler-side compensation, transactional session start, no pretending irreversible actions can roll back — is exactly the right shape. The tests were written honestly, too.

We're closing this PR for process reasons, not because of the idea: the branch is based on the July 10 main and now conflicts broadly (window sizing #52, hover latch, the new emulate/network tools all touched the same code), the review comments from July 18 haven't been addressed, and the Rust CI job on the head shows cancelled. The daemon-side piece (keeping the session queue busy until extension-side compensation completes) also still needs doing.

If you'd like to pick it up again, we'd genuinely welcome a rewrite on current main — drop the port-TOCTOU hunks (that fix already landed independently), port the signal threading onto the new dispatcher/manager code, and we're happy to re-review quickly. If we don't hear back and end up absorbing the core changes (dispatcher de-racing, transactional session start, snapshot dirty-write guard) into a maintainer PR, we'll credit you in the commit message. Either way — thank you for pushing this forward. 🙏

这个方向我们是认的：问题真实、方案正确。关闭只是因为 PR 已陈旧且需要按新 main 重写；欢迎随时回来重做，或由我们吸收核心思路并署名致谢。

NianJiuZst:
Rewritten from current main with the July review feedback addressed in #89. The replacement keeps the daemon session queue busy through extension-side compensation, adds the delayed-compensation E2E coverage, closes newly created fallback windows on cancellation, and surfaces cleanup failures explicitly. The obsolete port-TOCTOU hunks are not carried forward.

PR #19, opened 2026-07-10
fix(extension): harden websocket reconnect lifecycle

NianJiuZst:
## What problem this solves

The extension had two related connection-state races that could leave it apparently connected while no usable daemon channel existed.

At the transport layer, an unexpected close scheduled a reconnect timer. If keepalive or the user explicitly called `connect()` before that timer fired, WSTransport opened a replacement socket but did not cancel the pending timer. The timer later opened another socket. In addition, every socket callback wrote to shared state without checking whether that socket was still current. A delayed `close` from the old socket could therefore set `this.socket = null` after the new socket was already OPEN. The controller still displayed connected, while `send()` failed with "cannot send while not connected"; one of the extra sockets could also remain open after disconnect.

At the controller layer, handshake work was guarded by one global `handshakeInFlight` boolean. If a socket disconnected while its handshake was pending and a new socket connected, the new connection skipped its mandatory handshake. Conversely, a handshake error or timeout only changed the controller state: it left the physical transport connected, so keepalive saw an OPEN socket and never retried.

### Single-current-socket transport

WSTransport now assigns a monotonically increasing generation to every physical socket. Open, message, and close callbacks first verify both the socket object and generation before touching shared state. Events from a replaced socket are ignored.

Explicit `connect()` cancels a pending backoff timer before starting a new attempt. Reconnect timers call the same deduplicated connect path, and a failed physical attempt can create a new socket while preserving the original connect promise. `disconnect()` invalidates the current generation before closing the socket, preventing its synchronous or delayed close callback from restarting the reconnect loop.

### Per-connection handshake lifecycle

Each connected event now creates its own handshake generation and AbortController. A disconnect or replacement connection aborts the prior handshake immediately and removes its message listener instead of waiting for the ten-second timeout. Results are applied only when they still belong to the current connection generation.

If a current handshake fails, the controller actively disconnects the unusable physical socket and schedules a bounded retry. A protocol-version rejection remains a deliberate terminal disconnect; transient errors no longer leave transport and controller state disagreeing.

## User impact

Rapid disable/re-enable actions, keepalive ticks during backoff, delayed browser events, and service-worker reconnects no longer create multiple active daemon sockets. The popup state reflects the socket that `send()` will actually use. A fresh connection always performs its own handshake, and a handshake failure recovers instead of staying permanently stuck until the extension is manually toggled.

## Validation

  - 35 test files passed
  - 377 tests passed
- `corepack pnpm ext:build`
- `git diff --check`

New regression coverage verifies:

- explicit connect cancels pending backoff and does not create a third socket;
- delayed close events from an old socket cannot clear the current socket;
- outbound traffic continues through the replacement socket;
- a handshake AbortSignal stops the old waiter immediately;
- handshake failure disconnects the physical socket and retries;
- a reconnect starts a new handshake, while a late response for the prior generation is ignored.

## CI Baseline Compatibility

## CI Reliability Follow-up

### DIFF
--- a/apps/extension/src/lib/__tests__/connection-controller.test.ts
+++ b/apps/extension/src/lib/__tests__/connection-controller.test.ts
@@ -1,7 +1,7 @@
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { MIN_COMPATIBLE_PROTOCOL } from "../../transport/handshake";
-import type { ConnectionStateHandler, Transport } from "../../transport/transport";
-import type { ConnectionState, HandshakeResult } from "../../transport/types";
+import type { ConnectionStateHandler, FrameHandler, Transport } from "../../transport/transport";
+import type { ConnectionState, HandshakeResult, ProtocolFrame } from "../../transport/types";
 import { __testing__, ConnectionController } from "../connection-controller";
 
 vi.mock("../instance-id", () => ({
@@ -97,6 +97,7 @@ describe("computeConnectedState (protocol-based compat)", () => {
 function makeMockTransport(initialState: ConnectionState = "disconnected") {
   let state = initialState;
   const stateHandlers = new Set<ConnectionStateHandler>();
+  const messageHandlers = new Set<FrameHandler>();
   const transport = {
     get state() {
       return state;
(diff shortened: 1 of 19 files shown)


BB-fat:
One blocking issue remains in the handshake retry path: `performHandshake()` flattens structured RPC errors into a generic `Error`, so `version_too_old` is treated as transient and retried every second.

Please preserve the error code (for example, with a typed `HandshakeError`), treat version incompatibility as terminal, and make keepalive respect that terminal state. Add a fake-timer test proving that `version_too_old` does not trigger another `connect()` call. Please also update the branch and rerun the complete CI suite.

BB-fat:
Thank you @NianJiuZst — the websocket reconnect lifecycle hardening (handshake race fixes) still applies cleanly to main and covers a gap nobody else picked up. We verified locally that it merges with zero conflicts. Merging now. 🙏

感谢贡献！websocket 重连加固与 main 零冲突，直接合并。

BB-fat:
Closing in favor of #82, which carries these changes rebased onto current main (conflict with the session-disconnect cleanup from #17 resolved) — full credit to @NianJiuZst for the original work. Thank you!

已由 #82 以 rebase 后的形式合并，感谢贡献！

BB-fat:
Closing as merged via #82 — your reconnect-lifecycle hardening was rebased onto current main (which had gained the disconnect-cleanup hooks from #17 in the meantime), the two were composed, and it has landed. Thank you @NianJiuZst for the solid work! 🙏

本 PR 的内容已通过 #82（rebase 到最新 main 并与 #17 的断连清理融合）合并进主干，故关闭。感谢贡献！

PR #21, opened 2026-07-10
fix(update): apply staged Windows replacement

NianJiuZst:
## What Problem This Solves

On Windows, `bsk update` could report a successful update even though the installed executable was never replaced. The updater wrote a staged executable and a `.cmd` replacement script next to the running binary, then returned `InstallAction::Staged`, but nothing launched that script. The old executable therefore remained installed indefinitely.

This also created a daemon availability problem. The update flow stops a running daemon before replacing the binary. Because the staged replacement never ran—and because the Windows path did not preserve the previous daemon state—the command could leave users on the old version with their daemon stopped while still printing an update-style success message.

## How This Fix Works

- Launch the generated Windows replacement script as a detached, no-window `cmd.exe` process before the updater exits.
- Keep retrying the move until the currently running `bsk.exe` releases its file lock, then atomically replace the target from the staged executable.
- Carry the pre-update daemon state into the helper and restart the daemon only when it had been running before the update.
- Delete the helper script after a successful replacement so update artifacts do not accumulate beside the executable.
- Remove both staged files if the helper cannot be launched, and return an error instead of claiming that the update was staged successfully.
- Report staged updates with a distinct `staged` status and wording that accurately describes when the replacement will occur.
- Escape percent characters in generated paths so Windows environment-variable expansion cannot corrupt the helper commands.

## User Impact

Windows users can now run `bsk update` and have the staged version applied automatically after the command exits. A daemon that was running before the update comes back on the new binary, while an intentionally stopped daemon remains stopped. Failed helper launches are visible as failures rather than false-positive update reports.

## Validation

- `cargo fmt --check`
- Added focused tests covering replacement retries, conditional daemon restart, helper self-cleanup, and the no-restart path.

The full local Rust workspace suite passed. A local GNU Windows cross-check was also attempted, but this macOS environment does not provide the `x86_64-w64-mingw32-gcc` toolchain required by the transitive `aws-lc-sys` build; GitHub CI remains the authoritative clean-environment build check for the PR.

## CI Baseline Compatibility

## CI Reliability Follow-up

### DIFF
--- a/apps/extension/vitest.config.ts
+++ b/apps/extension/vitest.config.ts
@@ -21,9 +21,7 @@ export default defineConfig({
   },
   define: {
     __BSK_EXT_VERSION__: JSON.stringify(pkg.version),
-    __BSK_DAEMON_WS_URL__: JSON.stringify(
-      process.env.BSK_DAEMON_WS_URL ?? "ws://127.0.0.1:52800",
-    ),
+    __BSK_DAEMON_WS_URL__: JSON.stringify(process.env.BSK_DAEMON_WS_URL ?? "ws://127.0.0.1:52800"),
   },
   resolve: {
     alias: {
--- a/crates/bsk-cli/src/cli/update.rs
+++ b/crates/bsk-cli/src/cli/update.rs
@@ -5,6 +5,11 @@ use std::io::{Cursor, Read, Write};
 use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime};
 
+#[cfg(windows)]
+use std::os::windows::process::CommandExt;
+#[cfg(windows)]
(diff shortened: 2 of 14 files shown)


BB-fat:
Closing as merged via #85 — your staged-Windows-replacement fix was rebased onto current main (the redundant port-TOCTOU test hunks were dropped since that fix already landed independently; the core mechanism composed cleanly with the auto-update work from #77). Thank you @NianJiuZst — this completes the last mile of updates on Windows! 🙏

本 PR 的内容已通过 #85（rebase 到最新 main 并与 #77 自动升级组合）合并进主干，故关闭。感谢贡献！

PR #28, opened 2026-07-16
feat: add semantic browser recording via bsk record

shnpd:
Capture user flows as LLM-readable trace.json (v1) with intent, page context, and semantic targets. Wire record_start/stop/await through CLI, daemon, and extension; bump CLI to 0.1.8 and extension to 0.1.4.

### DIFF
--- a/apps/extension/src/content/ControlOverlay.tsx
+++ b/apps/extension/src/content/ControlOverlay.tsx
@@ -58,6 +58,7 @@ export function ControlOverlay({
       />
 
       <div
+        data-slot="control-overlay-blocker"
         onPointerDown={(event) => {
           if (automationBypass) return;
           event.preventDefault();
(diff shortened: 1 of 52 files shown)


BB-fat:
## Suggestion: Add a recording entry point in the Popup UI (generate a prompt for the Agent)

This PR’s recording capability (`bsk record`) is primarily meant to be invoked by an Agent via the CLI. There is currently no user-facing entry point in the Extension UI. BrowserSkill itself cannot call a large language model, and end users will not type CLI commands themselves.

Proposal: add a recording-related entry in the **Popup**, e.g. “Record a flow / Teach the Agent how to do this”:

1. The user clicks the entry in the Popup
2. The extension generates a **Prompt** that can be pasted directly to an Agent (with one-click copy)
3. The Prompt **must include the current BrowserID (`instance_id`)**, so the Agent can target the correct browser instance and run `bsk record start …`
4. The user pastes the Prompt to the Agent → the Agent starts recording → the user demonstrates in the Agent Window → clicks Finish → the Agent receives `trace.json`

This would turn the collaboration model—“humans only demonstrate; the Agent starts recording and saves the textbook”—from CLI-only documentation into an in-product discovery path, lowering first-time discovery and onboarding cost.

Could this be a follow-up to this PR, or a minimal Popup entry included in this PR?

BB-fat:
`RecordOverlay` doesn't match our existing overlay style (`ControlOverlay`):

- Dark card vs white pill + logo
- Red rectangular button vs orange pill CTA with icon
- No fade-in / breathe treatment

Please align with the existing bottom pill pattern before merge.

PR #29, opened 2026-07-17
Correct the incorrect command examples in the error message that does not indicate the mistake.

klren0312:
Correct the incorrect command examples in the error message that does not indicate the mistake.

image

The correct command should be `bsk session list`

image


BB-fat:
Thanks for catching this and sending the fix — much appreciated!

PR #32, opened 2026-07-21
fix(extension): #31 hide control overlay when screenshot

klren0312:
before change the screenshot has the animation border and control button
image

after change when did screenshot these were hidden
image

### DIFF
--- a/apps/extension/src/entrypoints/content.ts
+++ b/apps/extension/src/entrypoints/content.ts
@@ -17,10 +17,13 @@ import {
 } from "@/lib/help-bridge";
 import {
   isOverlayAgentOverlayResetMessage,
+  isOverlayScreenshotHideMessage,
   OVERLAY_AUTOMATION_BYPASS,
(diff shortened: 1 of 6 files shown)


BB-fat:
Hi @klren0312 — thank you for jumping on this so early! 🙏

The same underlying issue (#31, overlays appearing in screenshots) has since been addressed more comprehensively by #64, which is now merged: it hides **all** overlays (control mask, help/record/borrow panels) during capture, waits two animation frames so the compositor has actually produced an overlay-free frame before shooting, and covers both capture paths (full-page `captureVisibleTab` and CDP element clips).

Because of that, we're closing this PR as superseded — but your report and screenshots were genuinely useful in confirming the problem and shaping the fix. Thanks again, and we hope to see more contributions from you!

同样的问题已由 #64 更完整地修复并合并（截图期间隐藏全部 Overlay 并等待合成帧，覆盖两条截图路径），因此本 PR 作为已被覆盖项关闭。非常感谢你的报告和贡献！

PR #38, opened 2026-07-22
fix(install): verify release archive checksum before install

hobostay:
## Problem

The first-run installers (`install.sh`, `install.ps1`) downloaded and extracted the `bsk` release archive **with no integrity check**, even though the in-app `bsk update` path treats a sha256 checksum as mandatory:

> `crates/bsk-cli/src/cli/update.rs` — *"release … does not include a sha256 checksum; cannot safely auto-update"* and then `verify_sha256(...)`.

A compromised release artifact, CDN, or cache delivers a trojaned `bsk` binary and the installer installs it as-is. The release `version.json` already carries a per-asset `sha256` (`scripts/render-version-json.mjs` renders it with `--dist`), so the installers can verify exactly the way `bsk update` does.

## Change

Both installers now fetch `version.json` once (for both pinned `BSK_VERSION` and latest), read `assets[<platform>].sha256`, and verify the downloaded archive's sha256 **before** extracting:

- **`install.sh`** — POSIX `sed` parses the per-platform field (no `jq` dependency, matching the existing `sed`-based version parse); `sha256sum` / `shasum -a 256` computes the digest.
- **`install.ps1`** — `Invoke-RestMethod` parses the JSON and `Get-FileHash -Algorithm SHA256` computes the digest; comparison is case-insensitive (`-ieq`).

Failure handling:
- **Checksum mismatch → fatal** (fail-closed, matching `bsk update`).
- **Missing manifest / no published checksum / no sha256 tool → skip with a warning.** This keeps the first-run bootstrap working on minimal or offline machines instead of hard-failing a legitimate install.

This preserves install behavior while closing the integrity gap the updater already enforces.

## Validation

The installers are not exercised by CI (no shell/PowerShell test harness), so validated manually:
- `sh -n` / `bash -n` on `install.sh` (clean).
- Functional test of `extract_asset_sha256` against a realistic pretty-printed `version.json` (correct per-platform hash; empty → skip for an absent platform) and of `compute_sha256` against `sha256sum` (matching digest).
- `install.ps1` reviewed for PS 5.1 compatibility (`Get-FileHash`, dynamic member access `$manifest.assets.$platformKey.sha256`, `-ieq`).

`version.json` schema is unchanged (only consumed), so the `node-scripts` CI job is unaffected.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -51,26 +51,13 @@ function Get-PlatformTriple {
     }
 
     return @{
-        ArchId      = $archId
+        ArchId       = $archId
         TargetTriple = $windowsArch
+        PlatformKey  = "windows-$archId"
     }
 }
 
 # ── Version resolution ────────────────────────────────────────────────────────
-
-function Get-LatestVersion {
-    $versionJsonUrl = "${GitHub}/releases/latest/download/version.json"
-    Write-Log "fetching latest version from ${versionJsonUrl}"
-
-    try {
-        $version = (Invoke-RestMethod -Uri $versionJsonUrl).version
-        if (-not $version) { Write-Die "could not parse version from version.json" }
-        return $version
-    }
-    catch {
-        Write-Die "failed to fetch version.json: $_"
-    }
-}
 
 # ── PATH helpers ──────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


BB-fat:
Thank you @hobostay — this closes a real gap: the first-run installers now verify the release archive's sha256 (matching the guarantees `bsk update` already enforced), so a compromised artifact can't silently land on user machines. Clean work, and it lines up perfectly with the version.json checksum chain. Merging now. 🙏

感谢贡献！首装校验补上了发布链路的关键一环，直接合并。

PR #54, opened 2026-07-31
feat(session): support unfocused agent windows

### LINKED ISSUE #26: Feature Request: Add --no-focus / --minimized flag to bsk session start to prevent Agent Window from stealing focus
# Feature Request: Option to create Agent Window without stealing focus (minimized / background)

## Problem

Every time bsk session start is called, the Agent Window is created with focused: true in the extensions chrome.windows.create() call (see apps/extension/src/session-manager/agent-window.ts line 31). This causes the Agent Window to immediately pop to the foreground and steal keyboard/mouse focus away from whatever the user is currently working on.

For users who run browser automation tasks in the background while continuing to work in other applications (an IDE, a document editor, a video call, etc.), this forced focus-stealing is a significant disruption to their workflow.

## Proposed Solution

Add a CLI flag (and/or environment variable) to bsk session start that controls whether the Agent Window is created as a focused, visible window or as a minimized / unfocused window.

### Option A (recommended): --no-focus flag

bsk session start --no-focus

When this flag is set, the extension creates the Agent Window with focused: false in chrome.windows.create(). The window opens in the background without stealing focus from the users current active window. The user can manually bring it to the foreground when they need to observe what the agent is doing.

### Option B: --minimized flag

bsk session start --minimized

When set, the Agent Window is created in a minimized state. This is most useful for fully automated, fire-and-forget sessions where the user does not need to see the window at all during execution.

## Implementation Notes

In apps/extension/src/session-manager/agent-window.ts, the current code hardcodes focused: true. The focused parameter would need to be passed through the protocol from the CLI to daemon to extension. This requires:

1. Adding a field (e.g. focus: boolean) to the session start protocol message in crates/bsk-protocol
2. Passing it through the daemon to the extensions WebSocket message
3. Using it in agent-window.ts instead of the hardcoded true

(Optionally, for --minimized: use state: minimized in chrome.windows.create() instead.)

## Use Case

- A developer triggers bsk session start from their AI agent to scrape data from an internal dashboard. The Agent Window opens in the background, the data is collected, and the session ends without ever interrupting the developers focus.
- A user runs a nightly automated report collection via BSK. The Agent Window opens minimized, collects the data, and closes with zero user interaction needed.
- Any scenario where the user wants the agent to just work in the background without being visually intrusive.

## Workaround (for now)

Users can install an AutoHotkey script that detects new Edge/Chrome windows and minimizes them automatically, but this is fragile and requires third-party tools.

alectimison-maker:
## Summary

- add `bsk session start --no-focus` so Agent Windows can open without stealing focus
- carry an optional `focused` value through the CLI, daemon, protocol, and extension
- preserve the existing focused default when the field is omitted
- document the flag and cover both default and unfocused behavior at each boundary

Closes #26.

## Why

`bsk session start` currently hardcodes `focused: true` at the Chrome window boundary. That interrupts users who want an agent session to start while they continue working in another window.

The new flag is opt-in. Existing callers omit `focused`, and the extension continues to default it to `true`.

## Testing

- `corepack pnpm lint`
- `corepack pnpm --filter @browser-skill/extension test` (44 files, 497 tests)
- `corepack pnpm --filter @browser-skill/extension build`

## Compatibility

The protocol field is optional and omitted by default. Older clients therefore keep the current focused behavior, while `--no-focus` explicitly sends `focused: false`.

### DIFF
--- a/apps/extension/src/session-manager/__tests__/agent-window.test.ts
+++ b/apps/extension/src/session-manager/__tests__/agent-window.test.ts
@@ -44,3 +44,39 @@ describe("chromeAgentWindowApi.ensureActiveTab", () => {
     expect(update).not.toHaveBeenCalled();
   });
 });
+
+describe("chromeAgentWindowApi.create", () => {
+  const create = vi.fn();
+
+  beforeEach(() => {
+    vi.stubGlobal("chrome", {
+      windows: { create },
+    });
+    create.mockReset();
+    create.mockResolvedValue({ id: 100 });
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it("focuses Agent Windows by default", async () => {
+    await chromeAgentWindowApi.create(AGENT_WINDOW_HOME);
+
+    expect(create).toHaveBeenCalledWith({
+      type: "normal",
+      focused: true,
+      url: AGENT_WINDOW_HOME,
+    });
+  });
+
+  it("can create an Agent Window without stealing focus", async () => {
+    await chromeAgentWindowApi.create(AGENT_WINDOW_HOME, false);
+
+    expect(create).toHaveBeenCalledWith({
+      type: "normal",
+      focused: false,
+      url: AGENT_WINDOW_HOME,
+    });
+  });
+});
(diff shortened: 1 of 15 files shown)


BB-fat:
Closing as merged via #87 — we took over your design and reworked it onto current main so it composes with #52's window sizing: the `focused` flag now rides an options object (`create(url, { size, focused })` / `start(sessionId, { size, focused })`) instead of the colliding positional slot. Everything else (flag name, default-true semantics, four-layer plumbing, docs) follows your original approach. Thank you @alectimison-maker! 🙏

本 PR 的内容已通过 #87（按新 main 重做、与 #52 窗口尺寸共存）合并进主干，故关闭。感谢贡献！

PR #78, opened 2026-08-12
feat: add drag command and remote WS host binding support

icefairy:
```markdown
   ## Summary

   This PR adds two major features to BrowserSkill:

   ### 1. bsk drag command for sliders & drag-and-drop
   - New `bsk drag` command supporting three modes: element + delta, coordinate + delta, and absolute path
 trajectory
   - Human-like trajectory generation with randomized grab point for realistic interactions
   - Useful for slider controls, cross-origin iframe captchas (e.g. Aliyun nc_1_nocaptcha), and precise multi-stop
 paths
   - Includes unit tests and full protocol/CLI implementation

   ### 2. Remote WS host binding support (`0.0.0.0`)
   - New `--ws-host` flag and `BSK_DAEMON_WS_HOST` env var for binding the daemon WebSocket to a custom host
   - Enables remote access from outside the container/host machine
   - Updated daemon startup logic and IPC communication

   ### Files changed
   - **Extension:** `interaction.ts` (new drag handler), `dispatcher.ts`, `background.ts`, `types.ts`, tests
   - **CLI:** `interaction.rs` (new drag subcommand), `daemon.rs` (ws-host flag), `mod.rs`, `start.rs`
   - **Protocol:** `method.rs`, `tools/interaction.rs`
   - **Docs:** `SKILL.md` updated with drag command documentation
 ```

### DIFF
--- a/apps/extension/src/entrypoints/background.ts
+++ b/apps/extension/src/entrypoints/background.ts
@@ -69,10 +69,18 @@ export default defineBackground(() => {
         generation: overlayGeneration,
       };
     }
+    // Automation-only deployments (BSK_NO_OVERLAY=1 at build time) never
+    // show the on-page control mask: the orange stop button is useless
+    // in a headless browser and its capture layer swallows CDP mouse
+    // events (breaking drag automation).
+    const mode: OverlayMode =
+      typeof __BSK_NO_OVERLAY__ !== "undefined" && __BSK_NO_OVERLAY__
+        ? "hidden"
+        : (controlModes.get(ctx.sessionId) ?? "control");
     return {
       type: OVERLAY_AGENT_STATE,
       sessionId: ctx.sessionId,
-      mode: controlModes.get(ctx.sessionId) ?? "control",
+      mode,
       generation: overlayGeneration,
     };
   }
--- a/apps/extension/src/tools/__tests__/interaction.test.ts
+++ b/apps/extension/src/tools/__tests__/interaction.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it, vi } from "vitest";
 import { SessionManager } from "@/session-manager/manager";
 import type { CdpRunner } from "@/tools/shared";
 import {
+  buildHumanDragPath,
   handleClick,
   handleFill,
   handlePress,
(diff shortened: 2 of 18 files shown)


BB-fat:
Thanks @icefairy for this substantial contribution — the drag command design (three modes, human-like trajectories) covers slider/captcha scenarios nicely, and the full-chain wiring is thorough. Before we can merge, there are 4 blockers to fix:

🔴 1. **Flaky tests**: `buildHumanDragPath` leaves an unconverged ±3px random offset at the path end, but the test asserts the endpoint within ±1px — we reproduced ~55% failure rate locally (10k Monte Carlo runs), so frontend CI will go red. Same root cause lets the release point drift up to 3px from the target (slider captchas are sensitive to that) and makes `DragResult.to_x/to_y ≠ from+dx/dy`. Fix: zero/taper the offsets at both ends of the trajectory and land the release exactly on the target.

🔴 2. **Conflicts with main (mergeable: dirty)**: `crates/bsk-cli/src/daemon/start.rs` collides with the merged #77 (daemon auto-update). Needs a rebase.

🔴 3. **SKILL.md out of sync**: only `crates/bsk-cli/skill/SKILL.md` is updated; the root `skill/SKILL.md` is not (they are byte-identical on main). Also the PR's SKILL.md has a duplicated `### Observation` heading (lines ~167-168).

🔴 4. **Dangling docs**: SKILL.md documents `bsk screenshot --full-page` and types.ts adds `ScreenshotParams.full_page`, but nothing implements it (CLI/extension/protocol untouched) — looks like residue from another branch.

Smaller notes (🟡): `--ws-host 0.0.0.0` binds without auth and without warning (Origin check only accepts chrome-extension:// shapes) — please add a prominent warning in help/docs, and fail loudly instead of silently falling back to 127.0.0.1 when the host can't resolve; CLI uses `||` for coordinate-mode detection while the extension requires pairs, so a lone `--from-x` silently drops the target; `[v2-DRAG]` debug marker left in dispatcher error message; element-mode without dx/dy should be invalid_params; dump-schema.rs registration missing for DragParams/DragResult; consider splitting or documenting the BSK_NO_OVERLAY build flag.

Happy to re-review once these are addressed! 功能方向很好，修掉这 4 个卡点（尤其 flaky 测试和 rebase）后欢迎继续推进。

PR #89, opened 2026-08-13
fix(extension): make tool cancellation cooperative

NianJiuZst:
## Summary

This is a current-main rewrite of #18, based on `37fbaacc94a4e7c1b4198fc0a72756fe0b1e663a`.

- remove the dispatcher-level `Promise.race` that reported cancellation while the original handler was still running;
- acknowledge the separate `cancel` RPC immediately, while keeping the original RPC pending until handler work and compensation settle;
- thread `AbortSignal` through session startup, tab/window operations, observations, console/network reads, and emulation;
- make `session_start`, tab creation/borrowing, and fallback-window handling transactional where compensation is possible;
- keep the daemon session queue busy until the extension returns its final result or a bounded two-second cleanup timeout expires;
- surface compensation failures explicitly as `cleanup_failed`, including the orphan resource id.

The obsolete port-TOCTOU and CI-only commits from the original branch are intentionally not included because those changes already landed on `main`.

## Review feedback addressed

- **Queue ownership:** a forwarded cancellation no longer releases the per-session queue immediately. A second same-session RPC remains `session_busy` until the extension finishes cleanup; a wedged handler is bounded by `cancel_cleanup_timeout`.
- **Cooperative checks:** `tab_return` and multi-stage observation paths check cancellation after awaited steps and before later side effects or state publication.
- **Fallback compensation:** a fallback window created by `tab_return` is removed when cancellation wins before the tab move, and is also removed when an attempted move fails.
- **Visible cleanup failures:** failed cleanup is not swallowed. The final RPC returns `cleanup_failed` with the affected tab/window id, while retryable borrowed-tab state is retained.
- **Delayed-compensation E2E:** the Rust integration test gates the extension's final cancellation response, proves the session stays busy during that delay, then proves the queue reopens after cleanup.

## Cancellation boundary

Chrome operations that have already completed are not presented as rolled back. The guarantee is instead lifecycle honesty: the extension does not send the original response while its handler is still running invisibly, and the daemon does not release queue ownership before that response or the cleanup deadline.

For reversible allocations, cancellation performs compensation before the final response. In particular, cancelled or partially failed session startup removes its newly created Agent Window, and a cancelled observation cannot replace the session `RefStore` after a newer operation has moved on.

## Validation

- `corepack pnpm --filter @browser-skill/extension test` — 52 files, 647 tests passed
- `corepack pnpm lint`
- `corepack pnpm ext:build`
- `cargo test -p bsk --test cancel_forwarding --locked -- --nocapture` — 5 tests passed
- `git diff --check`

Replaces #18.

### DIFF
--- a/apps/extension/src/session-manager/__tests__/manager.test.ts
+++ b/apps/extension/src/session-manager/__tests__/manager.test.ts
@@ -75,6 +75,52 @@ describe("SessionManager", () => {
     await expect(sm.start("aa11")).rejects.toThrow(/already exists/);
   });
 
+  it("removes a newly created Agent Window when startup is aborted", async () => {
+    const aw = fakeAgentWindow();
+    let resolveCreate: (windowId: number) => void = () => {};
+    aw.createMock.mockImplementationOnce(
+      () =>
+        new Promise<number>((resolve) => {
+          resolveCreate = resolve;
+        }),
+    );
+    const sm = new SessionManager({ agentWindow: aw });
+    const controller = new AbortController();
+    const pending = sm.start("aa11", { signal: controller.signal });
+
+    controller.abort();
+    resolveCreate(777);
+
+    await expect(pending).rejects.toMatchObject({ name: "AbortError" });
+    expect(aw.removeMock).toHaveBeenCalledWith(777);
+    expect(sm.has("aa11")).toBe(false);
+  });
+
+  it("removes an incomplete Agent Window when active-tab setup fails", async () => {
+    const aw = fakeAgentWindow();
+    aw.ensureActiveTabMock.mockRejectedValueOnce(new Error("tab setup failed"));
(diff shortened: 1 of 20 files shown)


BB-fat:
Thank you @NianJiuZst for the quick rewrite on top of current main — this version is in great shape.

All four review points from 7/18 landed cleanly and with high quality: the queue now stays busy until compensation fully completes, cancellation checks are cooperative throughout, the fallback window is compensated as well, and `cleanup_failed` is now explicit. The gating in the delayed-compensation E2E test is especially solid.

We have two tiny follow-ups of our own (wrapping `stopAll`'s `remove` in try/catch, adding CLI error copy for `cleanup_failed` / `cancel_cleanup_timeout`, plus a one-line cancellation check in the hover latch) — no action needed from you; we'll pick those up separately.

Merging now. Contributions like this are very welcome — looking forward to the next one!

PR #96, opened 2026-08-17
fix(skill): preserve custom installed instructions

NianJiuZst:
## Summary

- record whether an installed `SKILL.md` came from the bundled source or `--source`
- hash managed bundled content so automatic sync only updates an untouched managed copy
- preserve custom, manually edited, and unknown marker-less installations
- expose preserved installations through daemon logs and `bsk doctor`

## Why

`bsk install-skill --source ...` installs user-owned instructions, but daemon startup currently rewrites every differing `SKILL.md` with the bundled copy. Starting a session can therefore silently destroy custom instructions.

The sidecar marker is written fail-safe: custom ownership is recorded before content replacement, and bundled markers include the exact installed-content hash. A marker-less file is only adopted automatically when it exactly matches the current bundle.

## Tests

- `cargo test -p bsk --locked`
- `cargo clippy -p bsk --all-targets --locked -- -D warnings`

### DIFF
--- a/crates/bsk-cli/src/skill_install/sync.rs
+++ b/crates/bsk-cli/src/skill_install/sync.rs
@@ -7,6 +7,41 @@ use super::{
     DEFAULT_SKILL_MD, SOURCE_CUSTOM, SOURCE_MARKER_FILE, bundled_source_marker, harness::HarnessId,
 };
 
+/// SHA-256 hex digests of `SKILL.md` content that `bsk` shipped in
+/// earlier releases, before `.bsk-source` markers existed. When a
+/// marker-less installed `SKILL.md` matches one of these
+/// fingerprints (or the current bundle), the daemon recognises it
+/// as a managed bundled install and rewrites it to the current
+/// bundle in the same sync pass. Without this list, anyone who
+/// installed before markers landed would be silently preserved as
+/// "unknown" forever, frozen on the version they first installed.
+const HISTORICAL_BUNDLED_FINGERPRINTS: &[&str] = &[
+    // Initial public release
+    "28d39215d1a6f658d55a41b8ecc2c6692bed628093824442927581f4a3e1bc5a",
+    // feat(record): add popup quick-actions launcher and make record --url optional
+    "875affb61f68c62712e038bb2a3a54efeaa20974aeab516aaaacd9605e25aa5a",
+    // fix(record): use example.com as the default record start URL
+    "c3d933f6390199b9a250e6fc66e441c0817fbceb16403d760606ff9bcd3c25ba",
+    // fix(runtime): fix early control return and update lifecycle
+    "735bd85400e725c33ebb1146bf55d1f997fd4e7f6b785da52fe600a5cd81d4ca",
+    // fix(vom): bug fix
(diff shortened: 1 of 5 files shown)


crates/bsk-cli/src/skill_install/sync.rs
```diff
@@ -54,6 +61,29 @@ fn sync_one(dest: &Path, source: &str) -> SyncOne {
...
+        .parent()
+        .expect("SKILL.md destination must have a parent")
+        .join(SOURCE_MARKER_FILE);
+    let managed_marker = match std::fs::read_to_string(&marker) {
+        Ok(value) if value == SOURCE_CUSTOM => return SyncOne::Preserved,
+        Ok(value) if value.starts_with("bundled:") => value,
+        Ok(_) => return SyncOne::Preserved,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+            if on_disk != source {
+                return SyncOne::Preserved;
```
iuyo5678: This  code line causes a side effect: users who have never upgraded to the latest version will never upgrade.

iuyo5678:
A very good PR, but we need to fix the bug where users who haven't upgraded to the latest version will never be able to upgrade automatically again, the cleanest solution is to prepare a historical fingerprint for the "no marker" case

NianJiuZst (commented):
Thanks for the catch. You're right — anyone who installed before `.bsk-source` markers existed would be frozen on their first-ever bundled `SKILL.md` because the new marker-aware sync refused to overwrite a differing file. Now fixed.

What changed in this update:

- Added `HISTORICAL_BUNDLED_FINGERPRINTS: &[&str]` in `crates/bsk-cli/src/skill_install/sync.rs` containing the SHA-256 hex digests of every `SKILL.md` ever shipped in this repo (initial public release through `feat(session): support unfocused agent windows --no-focus #87`, i.e. the same content as the current bundled `SKILL.md`).
- Extended the marker-less branch of `sync_one`: when there's no marker, hash the on-disk content and adopt as managed if it matches either the current bundle or any historical fingerprint. Matching the current bundle stamps a bundled marker and reports `UpToDate`; matching a historical fingerprint atomically rewrites to the current bundle, stamps a fresh bundled marker, and reports `Updated`. Anything else still goes to `Preserved` (unrecognised manual edits remain untouched).
- Refactored `sync_one` to share the atomic-replace + marker-refresh path between stale-managed upgrades and legacy-bundled adoptions (extracted into a small `atomic_replace` closure).
- Updated `SyncReport::preserved`'s docstring accordingly.
- New unit test `sync_adopts_markerless_legacy_bundle_and_upgrades_it` writes the real initial-release `SKILL.md` content (sha256 `28d39215...d4a3e1bc5a`, exactly matching the list entry) without a marker, syncs against a new bundle, and asserts the file was upgraded and a fresh bundled marker was stamped. The existing `sync_preserves_custom_edited_and_unknown_skills`, `sync_adopts_markerless_current_bundle_then_updates_it`, `sync_updates_outdated_skill`, etc. all still pass.

New head: `25af5c5` (rebased onto current `main@6b78211`).

Verified locally after the rebase:
- `cargo test -p bsk --locked --lib skill_install::sync` — all 7 tests pass, including the new one.
- `cargo clippy -p bsk --all-targets --locked -- -D warnings` — clean.

PR #98, opened 2026-08-17
fix(daemon): make idle disconnect updates atomic

NianJiuZst:
## Summary

- keep the last IPC activity timestamp and active connection count in one mutex-protected state
- update connection close count and close timestamp atomically
- recover poisoned activity-state locks consistently and avoid connection-count underflow
- add deterministic boundary tests for close/activity idle windows

## Why

The close callback previously decremented the atomic connection count before updating the separately locked activity timestamp. The idle task could observe the old expired timestamp between those operations, then see zero connections and shut the daemon down immediately after the last client disconnected instead of starting a fresh idle interval.

An idle check now observes a coherent snapshot: it sees either the still-open connection or the new close timestamp.

## Tests

- `cargo test -p bsk --locked`
- `cargo clippy -p bsk --all-targets --locked -- -D warnings`
- `cargo test -p bsk --test idle_exit --locked -- --test-threads=1`
- repeated the last-connection disconnect regression 3 additional times

### DIFF
--- a/crates/bsk-cli/src/daemon/start.rs
+++ b/crates/bsk-cli/src/daemon/start.rs
@@ -15,7 +15,6 @@ use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::path::Path;
 use std::sync::Arc;
 use std::sync::Mutex;
-use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::{Duration, Instant};
 
 use anyhow::{Context, Result};
@@ -291,8 +290,7 @@ pub fn run_foreground(cfg: DaemonConfig) -> Result<()> {
         });
 
         let started_at = Instant::now();
-        let activity: Arc<Mutex<Instant>> = Arc::new(Mutex::new(started_at));
-        let active_ipc_connections = Arc::new(AtomicUsize::new(0));
+        let activity = Arc::new(Mutex::new(IpcActivityState::new(started_at)));
         let (ipc_shutdown_tx, ipc_shutdown_rx) = tokio::sync::oneshot::channel::<()>();
         let status = ipc::DaemonStatus {
             started_at,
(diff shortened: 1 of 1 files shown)


crates/bsk-cli/src/daemon/start.rs
```diff
@@ -344,30 +338,24 @@ pub fn run_foreground(cfg: DaemonConfig) -> Result<()> {
...
-                    if last.elapsed() < daemon_idle {
+                    let ipc_is_idle =
+                        lock_activity(&activity).should_exit_for_idle(Instant::now(), daemon_idle);
+                    if !ipc_is_idle {
                         continue;
                     }
                     // Bridge from M2/M3 (which only knew about connection
                     // counts) to M4/M5 (browser + session registries):
                     // hold the daemon alive while any IPC client is
                     // connected, any browser is paired, or any session
```
iuyo5678: After the code was changed here, the comments no longer match the code, so please update this part.

iuyo5678:
The race condition here only exists in the nanosecond-scale gap between two adjacent statements, so it is almost impossible to observe in real situation. But consolidating the two states that change together into a single invariant will avoid the same kind of bug again, so it is still worthwhile. Once the code comments are updated to match the code, it can be merged.

NianJiuZst (commented):
Thanks for the review. Per your note, the idle-task comment now matches the post-refactor control flow: `should_exit_for_idle` already enforces the zero-connection + idle-window invariant under the single `IpcActivityState` lock, so the remaining `if` block only holds the daemon alive while any browser is paired or any session is live (design §3.2, M4/M5 registries).

New head: `8352c3c` (rebased onto current `main@6b78211`).

Verified locally after the rebase:
- `cargo test -p bsk --locked --lib -- ipc_close ipc_activity` — both new unit tests pass.
- `cargo clippy -p bsk --all-targets --locked -- -D warnings` — clean.
- `cargo test -p bsk --test idle_exit --locked -- --test-threads=1` — the integration tests need an actual daemon process binding a UDS in `/var/folders/...`; on this machine that bind is rejected by the sandbox (`Operation not permitted (os error 1)`), so I could not re-run them locally. The PR did not touch those tests; they should still be green on your CI.

No logic change, only the comment refresh.

PR #99, opened 2026-08-17
fix(record): preserve committed empty fill steps

### LINKED ISSUE #58: record: committed input clears are unconditionally omitted from exported traces
## Summary

When a user clears a text input that previously contained a value and then
commits the change, the exported trace does not contain an empty `fill` step.

As a result, some workflows that depend on removing an existing value cannot
be reproduced from the trace.

## Steps to reproduce

Use a page containing a prefilled input and a button:

```html
<input aria-label="Search query" value="browser" />
<button>Apply</button>
```

1. Start `bsk record` on the page.
2. Clear the prefilled input.
3. Click **Apply**.
4. Finish recording and inspect `trace.json`.

## Actual behavior

The Apply click is recorded, but the input clear is omitted:

```json
{
  "steps": [
    {
      "op": "click",
      "target": {
        "role": "button",
        "name": "Apply",
        "tag": "button"
      }
    }
  ]
}
```

## Expected behavior

The committed transition from the existing value to an empty value should be
represented before the click:

```json
{
  "steps": [
    {
      "op": "fill",
      "target": {
        "role": "textbox",
        "name": "Search query",
        "tag": "input"
      },
      "value": ""
    },
    {
      "op": "click",
      "target": {
        "role": "button",
        "name": "Apply",
        "tag": "button"
      }
    }
  ]
}
```

This report is about a committed change from a non-empty value to an empty
value. A transient empty state while the user continues editing does not need
to become a separate step.

## Source observation

The capture layer detects changes relative to the field's initial value, but
the trace reducer currently excludes every non-redacted blank `fill`:

```ts
if (step.op === "fill" && !(step.value ?? "").trim() && !step.redacted) return false;
```

This appears to make workflows that intentionally clear an existing value
impossible to express in the exported trace.

If preserving committed empty values aligns with the intended recording semantics, I'd be happy to work on the change.

shnpd:
## Summary
- preserve committed empty fill steps in exported recording traces
- add a regression test for intentionally cleared input values

Closes #58

## Test plan
- [x] `pnpm ext:test` (648 tests)
- [x] `pnpm --filter @browser-skill/extension compile`
- [x] `pnpm lint`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/extension/src/lib/__tests__/trace-reducer.test.ts
+++ b/apps/extension/src/lib/__tests__/trace-reducer.test.ts
@@ -104,6 +104,27 @@ describe("reduceTraceSteps", () => {
     });
   });
 
+  it("keeps committed empty fill steps", () => {
+    const { steps } = reduceTraceSteps(
+      [
+        {
+          op: "fill",
+          target: { tag: "input", role: "textbox", name: "Search query" },
+          value: "",
+          page_url: "https://example.com/search",
+        },
+      ],
+      "https://example.com/search",
+    );
+
+    expect(steps).toEqual([
+      expect.objectContaining({
+        op: "fill",
+        value: "",
+      }),
+    ]);
+  });
+
   it("maps select navigated_to onto effect.navigated_to (page id)", () => {
     const { pages, steps } = reduceTraceSteps(
       [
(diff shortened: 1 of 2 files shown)


iuyo5678:
LGTM

PR #126, opened 2026-08-21
feat(interaction): add explicit focus and blur primitives

NianJiuZst:
## Summary

- add explicit `bsk focus` and `bsk blur` commands with ref/selector targeting
- route both methods through the daemon and Agent Window sandbox, including OOPIF refs
- verify the target's deep active-element state and return machine-readable focus results
- add protocol schemas, CLI parsing, IPC coverage, extension tests, and skill documentation

## Motivation

Focus was previously only an incidental part of `fill`, `press`, and `select`. Agents could not deliberately enter or leave focus-driven UI states without falling back to arbitrary JavaScript.

## Validation

- `pnpm lint`
- extension TypeScript compile
- extension tests: 57 files / 690 tests
- extension production build

### Real browser probe

Loaded this branch's unpacked extension into Chrome for Testing 152 with an isolated daemon/profile:

1. `bsk focus '#probe'` returned `focused: true`; an independent page evaluation observed `document.activeElement.id === "probe"`.
2. `bsk blur '#probe'` returned `was_focused: true, focused: false`; an independent page evaluation observed that the target was no longer `document.activeElement`.


apps/extension/src/tools/interaction.ts
```diff
@@ -215,6 +218,157 @@ async function resolveBackendNode(
...
+  const node = await resolveBackendNode(deps.cdp, ctx, target, params, "blur");
+  if (isRpcError(node)) return node;
+  const nodeCdp = cdpRunnerForTarget(deps.cdp, node.cdpTarget);
+  deps.cdp.trackSessionTab?.(ctx.sessionId, target.tabId);
+  const abortedBeforeBlur = throwIfAborted(deps.signal);
+  if (abortedBeforeBlur) return abortedBeforeBlur;
+
+  try {
+    const objectIdOrErr = await backendNodeToObject(nodeCdp, target.tabId, node.backendNodeId);
+    if (isRpcError(objectIdOrErr)) return objectIdOrErr;
```
iuyo5678: The problem probability is relatively low, but I think it's worth fixing, because the cost of fixing is low.

Could we re-check throwIfAborted(deps.signal) here before calling Runtime.callFunctionOn? Cancellation may arrive while backendNodeToObject() is awaiting DOM.resolveNode. In that case, the earlier check has already passed, and this.blur() still executes despite cancellation, potentially triggering page-side validation or autosave. Returning the cancellation error here would prevent that follow-up mutation.

apps/extension/src/tools/interaction.ts
```diff
@@ -215,6 +218,157 @@ async function resolveBackendNode(
...
+
+const DEEP_FOCUS_CHECK = `function() {
+  const deepActiveElement = (root) => {
+    let active = root && root.activeElement;
+    while (active && active.shadowRoot && active.shadowRoot.activeElement) {
+      active = active.shadowRoot.activeElement;
+    }
+    return active;
+  };
+  return { focused: deepActiveElement(this.ownerDocument) === this };
```
iuyo5678: Could we start the focus check from this.getRootNode() instead of this.ownerDocument? For an input inside a closed shadow root, document.activeElement points to the host, but host.shadowRoot is null. The traversal therefore stops at the host and reports failure even when the input is actually focused.
The same issue affects both checks in BLUR_TARGET: was_focused can incorrectly be false, and blur can report success even if a blur handler refocuses the input. Could we update all three checks and add regression tests for closed shadow roots, including refocusing during blur?

iuyo5678:
To advance the merge, I will add the fixes from the review in this branch and resolve conflicts with the main branch. After verification, I will proceed with the merge. Thanks to the original implementation.

PR #127, opened 2026-08-21
feat(interaction): add scroll-to-element primitive

NianJiuZst:
## Summary

- add `bsk scroll-to <ref-or-selector>` backed by the new `tool.scroll_to` RPC
- resolve snapshot refs and selectors inside the Agent Window, including OOPIF targets, and scroll both frame owners and the target with Chromium CDP
- return the target's visible top-level viewport bounds and integrate dialogs, cancellation, hover release, and the interrupt gate

## Validation

- `pnpm lint`
- extension TypeScript compile, 58 test files / 688 tests, and production build
- real Chromium probe with the branch-built CLI and unpacked extension in an isolated daemon/profile:
  - before: `scrollY=0`, target `top=2508`, viewport `720`
  - after `bsk scroll-to '#target'`: `scrollY=2178`, target `top=330`, `bottom=390`, independently verified `visible=true`


iuyo5678:
To advance the merge, I will add the fixes from the review in this branch and resolve conflicts with the main branch. After verification, I will proceed with the merge. Thanks to the original implementation.

PR #151, opened 2026-08-28
docs(skill): clarify bsk evaluate exit-code vs --json .ok semantics

xiyanjun:
## What

Addresses the discoverability gap raised in #133.

`bsk evaluate`'s exit code reflects RPC success only, not JS success. The structured `--json` output already returns `ok` / `value` / `error`, but SKILL.md only said "JS throw → stderr, exit 0", which led agents to misread `exit 0` as "JS succeeded".

## Change

Two small doc clarifications, applied to both `skill/SKILL.md` and `crates/bsk-cli/skill/SKILL.md` (kept in sync):

- Command reference (`Scripting & timing`): note that programmatic callers should use `--json` and check `.ok` to detect JS errors.
- Exit-code table (`0` row): point `evaluate` callers to `--json` `.ok` to distinguish JS success from RPC success.

Documentation only — no behavior change.

### DIFF
--- a/crates/bsk-cli/skill/SKILL.md
+++ b/crates/bsk-cli/skill/SKILL.md
@@ -210,7 +210,7 @@ Both capture from the moment the tab is attached and read a bounded per-tab buff
 
 | Command | Summary |
 |---------|---------|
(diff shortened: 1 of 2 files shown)


iuyo5678 (approved):
LGTM

PR #154, opened 2026-08-28
fix: let request-help return structured timeout outcomes

giattijunior:
## Problem
`request-help` passes its user-facing timeout to the generic per-session queue. When no human action occurs, the extension's timer and the daemon queue timeout race at the same deadline; the queue can return a generic RPC timeout before the extension returns the documented `timed_out` result.

## Fix
Give only `tool.request_help` one second of daemon-side response grace. The original `timeout_ms` remains unchanged in the extension-facing params, so the user-visible deadline is preserved. Other tools keep their existing timeout behavior.

## Validation
- `git diff --check` passes.
- Reproduced on CLI `0.1.11`: `bsk request-help --timeout 5s` returned generic RPC timeout.
- Confirmed the structured completion path with valid `completion-criteria`.
- Cargo tests could not be run in the validation environment because Rust/Cargo is not installed.

### DIFF
--- a/crates/bsk-cli/src/daemon/ipc.rs
+++ b/crates/bsk-cli/src/daemon/ipc.rs
@@ -328,6 +328,7 @@ async fn handle_tool_dispatch(
         Ok(timeout) => timeout,
         Err(err) => return ResponseBody::Err(err),
     };
+    let dispatch_timeout = tool_queue_timeout(method, timeout);
     let inflight_guard = match state.tool_inflight.register(cli_rpc_id, session_id.clone()) {
         Ok(guard) => guard,
         Err(err) => {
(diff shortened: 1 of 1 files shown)


crates/bsk-cli/src/daemon/ipc.rs
```diff
@@ -328,6 +328,7 @@ async fn handle_tool_dispatch(
         Ok(timeout) => timeout,
         Err(err) => return ResponseBody::Err(err),
     };
+    let dispatch_timeout = tool_queue_timeout(method, timeout);
```
iuyo5678: some problem with this code，the method parameter should be borrowed in rust， it won't compile right。

PR #161, opened 2026-08-30
feat(i18n): add Korean localization

dr-coton:
## Summary

Korean browser locales currently fall back to English. Add `ko-KR` translations for the shared and extension namespaces and register them with the existing resource-driven language detection. Align Korean protocol-version guidance with the connected, upgrade-recommended state, preserve the current `cliProtocol` interpolation, and clarify that help requests wait for the user to complete a step.

Normalize persisted language tags through the same resource-based mapping on both initial storage reads and cross-context changes. A saved `ko` now selects `ko-KR` without overriding an existing language preference or generating echo writes. Add regression coverage for actual resource loading, fallback behavior, storage synchronization, translation keys/interpolation, and Korean popup recording instructions. Run localization tests in CI and document language selection and translation maintenance.

## Validation

- `pnpm --filter @browser-skill/i18n test` — 52 passed.
- `pnpm ext:test` — 1,198 passed; 40 existing opt-in browser tests skipped.
- `pnpm --filter @browser-skill/extension exec wxt prepare` and `pnpm --filter @browser-skill/extension compile` — passed.
- `pnpm ext:build` — passed.
- Biome on tracked/new project files, Stylelint, DSH plugin typecheck and DSH plugin tests — passed (232 plugin tests). Lint components were run separately because an existing ignored local worktree contains a nested Biome root configuration that blocks the root `pnpm lint` scan.
- Visual inspection of production bundles in isolated headless Chrome with mocked extension APIs: Korean popup and recording views at 320/340 px, borrow confirmation, and full/compact help overlays. No horizontal overflow or unresolved interpolation observed in the popup.


packages/i18n/src/i18n.ts
```diff
@@ -17,6 +19,10 @@ const resources = {
     common: enUSCommon,
     extension: enUSExtension,
   },
+  "ko-KR": {
```
iuyo5678: add  a  “ko”  match ，in  [chrome docs](https://developer.chrome.com/docs/extensions/reference/api/i18n?hl=zh-cn#locales) ，"ko" also represents Korean.

PR #163, opened 2026-09-01
fix(extension): support UUID generation on insecure origins

dayi:
## Summary

Fix recording initialization on ordinary HTTP intranet pages where
`crypto.randomUUID()` is unavailable.

## Changes

- Add a UUID v4 helper using `crypto.getRandomValues()` as fallback.
- Replace all three extension-side `crypto.randomUUID()` call sites.
- Add tests for native and fallback UUID generation.

## Validation

- 74 test files passed
- 800 tests passed
- TypeScript compilation passed
- WXT production build passed
- Biome check passed

### DIFF
--- a/apps/extension/src/content/__tests__/record-frame-agent.test.ts
+++ b/apps/extension/src/content/__tests__/record-frame-agent.test.ts
@@ -45,8 +45,30 @@ function portHarness() {
 
 describe("RecordFrameAgent", () => {
   afterEach(() => {
+    vi.unstubAllGlobals();
     document.documentElement.removeAttribute(RECORD_DOCUMENT_ATTRIBUTE);
     document.body.replaceChildren();
+  });
+
+  it("starts when crypto.randomUUID is unavailable", async () => {
+    const harness = portHarness();
+    const getRandomValues = vi.fn((bytes: Uint8Array) => {
+      bytes.fill(0x11);
+      return bytes;
+    });
+    vi.stubGlobal("crypto", { getRandomValues });
+    vi.stubGlobal("chrome", {
+      runtime: {
+        connect: vi.fn(() => harness.port),
+        sendMessage: vi.fn(),
+      },
+    });
+    const agent = new RecordFrameAgent();
+
+    await expect(
+      agent.start({ type: RECORD_FRAME_START, requestId: "rec-http", startedAtMs: 10 }),
+    ).resolves.toEqual({ ok: true });
+    expect(getRandomValues).toHaveBeenCalled();
   });
 
   it("keeps a failed stop retryable and flushes the final dirty fill", async () => {
--- a/apps/extension/src/content/record-step-delivery.ts
+++ b/apps/extension/src/content/record-step-delivery.ts
@@ -1,3 +1,4 @@
+import { createRandomUuid } from "@/lib/random-uuid";
 import {
   isAcceptedRecordStepAck,
   RECORD_STEP,
@@ -18,7 +19,7 @@ export class RecordStepDelivery {
   constructor(
     requestId: string,
     send: SendRecordStepMessage = (message) => chrome.runtime.sendMessage(message),
-    producerId: string = crypto.randomUUID(),
+    producerId: string = createRandomUuid(),
   ) {
     this.#requestId = requestId;
     this.#producerId = producerId;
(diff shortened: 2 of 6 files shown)


iuyo5678:
Add a regression test for RecordFrameAgent——simulate crypto.randomUUID being unavailable, then verify that recording can still start successfully.
This can prevent the fixed business failure path from reappearing?

dayi:
> Add a regression test for RecordFrameAgent——simulate crypto.randomUUID being unavailable, then verify that recording can still start successfully. This can prevent the fixed business failure path from reappearing?


BrowserSkill fails to start recording on enterprise intranet websites served over HTTP, for example:

`http://10.x.x.x:9000/`

Running bsk record start fails with:

`failed to start recording in content script`

The browser console shows the actual error:

`TypeError: crypto.randomUUID is not a function
    at .../record-frame.js `

The recording content script calls crypto.randomUUID() during initialization. This API may not be available when running on non-secure HTTP pages, causing the recording process to fail before it can start.

This is common in enterprise environments, where many internal OA, ERP, administration, and legacy systems are still deployed over HTTP.

Therefore, this issue prevents BrowserSkill from recording a significant class of enterprise intranet applications and should be addressed as a recording compatibility issue rather than a site-specific problem.

iuyo5678:
> > Add a regression test for RecordFrameAgent——simulate crypto.randomUUID being unavailable, then verify that recording can still start successfully. This can prevent the fixed business failure path from reappearing?
> 
> BrowserSkill fails to start recording on enterprise intranet websites served over HTTP, for example:
> 
> `http://10.x.x.x:9000/`
> 
> Running bsk record start fails with:
> 
> `failed to start recording in content script`
> 
> The browser console shows the actual error:
> 
> `TypeError: crypto.randomUUID is not a function at .../record-frame.js `
> 
> The recording content script calls crypto.randomUUID() during initialization. This API may not be available when running on non-secure HTTP pages, causing the recording process to fail before it can start.
> 
> This is common in enterprise environments, where many internal OA, ERP, administration, and legacy systems are still deployed over HTTP.
> 
> Therefore, this issue prevents BrowserSkill from recording a significant class of enterprise intranet applications and should be addressed as a recording compatibility issue rather than a site-specific problem.

You misunderstood what I meant.
I agree with the problem context and the fallback direction. 
The main focus here is the level of test coverage: the newly added tests currently only cover createRandomUuid(), and do not cover the RecordFrameAgent startup path where the actual failure occurred this time. I suggest simulating crypto.randomUUID being unavailable in the existing record-frame-agent.test.ts, and asserting that agent.start() returns { ok: true }. This is the only way to prevent the helper unit test from still passing if a direct call is reintroduced later.

dayi:
> > > Add a regression test for RecordFrameAgent——simulate crypto.randomUUID being unavailable, then verify that recording can still start successfully. This can prevent the fixed business failure path from reappearing?
> > 
> > 
> > BrowserSkill fails to start recording on enterprise intranet websites served over HTTP, for example:
> > `http://10.x.x.x:9000/`
> > Running bsk record start fails with:
> > `failed to start recording in content script`
> > The browser console shows the actual error:
> > `TypeError: crypto.randomUUID is not a function at .../record-frame.js `
> > The recording content script calls crypto.randomUUID() during initialization. This API may not be available when running on non-secure HTTP pages, causing the recording process to fail before it can start.
> > This is common in enterprise environments, where many internal OA, ERP, administration, and legacy systems are still deployed over HTTP.
> > Therefore, this issue prevents BrowserSkill from recording a significant class of enterprise intranet applications and should be addressed as a recording compatibility issue rather than a site-specific problem.
> 
> You misunderstood what I meant. I agree with the problem context and the fallback direction. The main focus here is the level of test coverage: the newly added tests currently only cover createRandomUuid(), and do not cover the RecordFrameAgent startup path where the actual failure occurred this time. I suggest simulating crypto.randomUUID being unavailable in the existing record-frame-agent.test.ts, and asserting that agent.start() returns { ok: true }. This is the only way to prevent the helper unit test from still passing if a direct call is reintroduced later.

Thanks, that makes sense.

The helper unit test only verifies the fallback itself, but does not protect the actual recording startup path where this issue occurred.

I'll add a regression test in `record-frame-agent.test.ts` that simulates `crypto.randomUUID` being unavailable and verifies that `agent.start()` still succeeds. This should also catch any future reintroduction of direct `crypto.randomUUID()` calls in the recording startup path.

iuyo5678 (approved):

Issue #228, opened 2026-09-11
[`@wxg-prc-cpg/browser-skill-dsh-plugin`] 0.2.1 is incompatible with `dsh-better-sidebar` 0.19.0 — sidebar render crash and an inert tab path opened as a real file


kiterunner1:
## Summary

`@wxg-prc-cpg/browser-skill-dsh-plugin@0.2.1` (npm `latest`) is broken when used together with `dsh-better-sidebar@0.19.0`, the sidebar version required by DSH `0.1.5-rc.1`. There are two independent defects, both in the plugin's sidebar integration:

1. **Sidebar render crash** on session switch or page refresh:
   `TypeError: Cannot read properties of undefined (reading 'kind')`.
2. **The live-view tab opens a nonexistent file.** The auto-opened tab carries the fake path `browser-skill:observation`; on 0.19.0 that path is resolved as a real file, so the tab renders a `cannot resolve target … ENOENT` error instead of the browser view.

Both are reproducible with a default installation, on the versions the README pairs with DSH `0.1.5-rc.1`. No newer npm release is available.

## Environment

| Component | Version |
| --- | --- |
| `@deepseek-ai/dsh` | `0.1.5-rc.1` |
| `@wxg-prc-cpg/browser-skill-dsh-plugin` | `0.2.1` (npm `latest`) |
| `dsh-better-sidebar` (`omdsh-dev/DSH-better-sidebar`) | `0.19.0` |
| OS | Windows 10 / 11 |

## Steps to reproduce

1. Install into the `web` profile: DSH `0.1.5-rc.1`, `dsh-better-sidebar@0.19.0`, `@wxg-prc-cpg/browser-skill-dsh-plugin@0.2.1`.
2. Start `dsh --profile web` and open the Web UI.
3. Start a BrowserSkill session, then switch sessions (or reload the page) → defect 1: the browser console shows the `TypeError`, and the sidebar reports `[dsh-better-sidebar] render error: TypeError: …`.
4. Open the right sidebar's **Files** page → defect 2: a tab whose title is the path itself (`browser-skillobservation`) with a red `cannot resolve target … ENOENT` panel. Closing it does not help: it is re-created while a visible browser session exists.

## Defect 1 — sidebar render crash

```
TypeError: Cannot read properties of undefined (reading 'kind')
    at leafNodes (client.js:4463:1)
    at observationTabOpen (client.js:4472:1)
    at evaluate (client.js:4507:1)
[...]

PR #231, opened 2026-09-11
fix(client): guard observationTabOpen against a missing right-column tree (DSH 0.1.5 / better-sidebar 0.19+)

hZsFN:
## Summary

`observationTabOpen()` crashes whenever the sidebar snapshot exists but carries no right-column tree — which is the normal shape on **DSH 0.1.5 with better-sidebar 0.19+**, so the observation tab probe throws on every render.

## Root cause

DSH 0.1.5 moved the right column to dsh's native Sidebar. better-sidebar 0.19.x therefore no longer emits `splits` in its sidebar state (`makeDefaultState()` now returns only `bottomSplits`), but this plugin still declares it as required and walks it:

```ts
export interface SidebarStateLike {
  splits: SidebarNodeLike;        // declared required — absent at runtime on 0.1.5+
  bottomSplits: SidebarNodeLike;
  panelOpen?: boolean;
}

export function observationTabOpen(state: SidebarStateLike | undefined): boolean {
  if (state === undefined) return false;
  for (const root of [state.splits, state.bottomSplits]) {   // state.splits === undefined
    for (const leaf of leafNodes(root)) {                    // leafNodes(undefined)
```

`leafNodes()` dereferences `node.kind` on entry, so the first iteration throws:

```
    at leafNodes (...)
```

## User-visible impact

The observation tab is registered inside better-sidebar's render tree, so the throw is contained by *its* error boundary and rendered as a persistent red banner in the bottom-left corner of the UI:

```
dsh-better-sidebar: Cannot read properties of undefined (reading 'kind')   [重试]
```

It appears on every page load and misattributes the fault to `dsh-better-sidebar`, which sends users chasing the wrong plugin. The "is my observation tab already open?" probe also fails every time it runs.

## Fix

- Mark `splits` optional — it genuinely is absent on 0.1.5+ while earlier versions still provide it.
- Skip absent roots before walking.

```diff
 export interface SidebarStateLike {
-  splits: SidebarNodeLike;
+  splits?: SidebarNodeLike;
   bottomSplits: SidebarNodeLike;
```

```diff
   for (const root of [state.splits, state.bottomSplits]) {
+    if (root === undefined) continue;
     for (const leaf of leafNodes(root)) {
```

## Verification

Reproduced and verified against a live install (Windows, dsh `0.1.5-rc.2` + `dsh-better-sidebar` `0.19.0-alpha.1`, pulled in through `@linxin666/dsh-web-all` `0.3.20`):

- Client console captured with a headless Chromium session against the running UI.
  - **Before:** `TypeError: Cannot read properties of undefined (reading 'kind') at leafNodes` plus the red banner.
  - **After** applying this same guard to the built `lib/client.cjs`: no console TypeError, no banner, and observation-tab detection still behaves correctly (re-opening an already-open tab is suppressed as before).
- `node --check` on the patched bundle passes.

Happy to reshape the fix if you'd rather keep `splits` required and only guard at the call site — the type change is included because the field is genuinely optional on current DSH.

### DIFF
--- a/packages/dsh-plugin-browserskill/src/client/observation-sidebar.tsx
+++ b/packages/dsh-plugin-browserskill/src/client/observation-sidebar.tsx
@@ -64,7 +64,12 @@ export interface SidebarSplitLike {
 export type SidebarNodeLike = SidebarLeafLike | SidebarSplitLike;
 
 export interface SidebarStateLike {
-  splits: SidebarNodeLike;
+  /**
+   * Right-column split tree. DSH 0.1.5 (better-sidebar 0.19+) moved the right
+   * column to dsh's native Sidebar and no longer emits this field, so it is
+   * optional here; earlier versions always provide it.
+   */
+  splits?: SidebarNodeLike;
   bottomSplits: SidebarNodeLike;
   /** Whether the right panel is expanded (the merged drawer on narrow screens). */
   panelOpen?: boolean;
@@ -170,7 +175,11 @@ function* leafNodes(node: SidebarNodeLike): Generator<SidebarLeafLike> {
 /** Whether a tab of our type is already open in either sidebar workbench. */
 export function observationTabOpen(state: SidebarStateLike | undefined): boolean {
   if (state === undefined) return false;
+  // A root may be absent even when the snapshot exists: `splits` is gone on
+  // DSH 0.1.5+ (better-sidebar 0.19+), where the right column belongs to dsh's
+  // native Sidebar. Guard before walking so we do not dereference undefined.
   for (const root of [state.splits, state.bottomSplits]) {
+    if (root === undefined) continue;
     for (const leaf of leafNodes(root)) {
       if (leaf.tabs.some((tab) => tab.type === OBSERVATION_TAB_TYPE)) return true;
     }
--- a/packages/dsh-plugin-browserskill/tests/client/observation-sidebar.test.tsx
+++ b/packages/dsh-plugin-browserskill/tests/client/observation-sidebar.test.tsx
@@ -117,7 +117,7 @@ function withTabOpened(state: SidebarStateLike, type: string): SidebarStateLike 
     }
     return { ...node, children: node.children.map(walk) };
   };
-  return { ...state, splits: walk(state.splits) };
+  return { ...state, splits: state.splits === undefined ? undefined : walk(state.splits) };
 }
 
 function makeSidebar(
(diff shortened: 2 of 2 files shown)


iuyo5678:
Frontend CI failed in the dsh-plugin typecheck inside pnpm lint (the job dies in ~18s, so tests/build never run):
```
tests/client/observation-sidebar.test.tsx(120,35): error TS2345:
Argument of type 'SidebarNodeLike | undefined' is not assignable to parameter of type 'SidebarNodeLike'.
```
Making SidebarStateLike.splits optional broke the test helper withTabOpened:
```
return { ...state, splits: walk(state.splits) };
```
walk still requires SidebarNodeLike, but state.splits is now possibly undefined. Tightening that call site should go green — e.g. walk bottomSplits when splits is missing, or narrow the type before calling walk.

A regression test would also help: observationTabOpen on a snapshot with no splits and only bottomSplits should not throw, and should still return true when the tab lives in the bottom tree. That’s the crash this PR is fixing; the existing tests all still assume both trees are present.

hZsFN:
Hi maintainers 👋 — I pushed a follow-up commit (7bf0cdf) addressing the typecheck failure from the previous CI run: `withTabOpened` now keeps a missing `splits` as-is (instead of feeding `undefined` to the tree walker), and I added a regression test covering `observationTabOpen` when the right-column tree is absent — the DSH 0.1.5 / better-sidebar 0.19+ case this PR fixes. The new run is waiting for approval (action_required), so could you approve the workflow run when you get a chance? Thanks!

devicefloat-svg:
Independent confirmation on the **newest** sidebar release, from a Windows setup 鈥?this is not limited to 0.19.0:

| Component | Version |
| --- | --- |
| `@deepseek-ai/dsh` | `0.1.5-rc.1` (web profile) |
| `dsh-better-sidebar` | **`0.19.1`** |
| `@wxg-prc-cpg/browser-skill-dsh-plugin` | `0.2.1` (npm `latest`) |

Same stack, and it fires on **every** sidebar state publish 鈥?session switch, page refresh, and the plugin's own first `evaluate()` during registration 鈥?not only on session switches:

```
    at leafNodes 鈫?observationTabOpen 鈫?evaluate 鈫?SidebarStore.publish 鈫?apply
    at componentDidCatch 鈫?[dsh-better-sidebar] render error: 鈥?```

We are running exactly this guard (plus two further local stopgaps for the second defect in #228: dropping the fake `path` seed and wrapping the auto-open in `try`/`catch` for the "no session surface is mounted" throw). With all three the page console is completely clean and the observation tab renders the live view again.

+1 to landing this 鈥?npm `latest` is still `0.2.1`, so every consumer has to patch the shipped bundle today.

PR #235, opened 2026-09-11
fix(dsh-plugin): tolerate an unpopulated sidebar layout on first paint

### LINKED ISSUE #232: [dsh plugin 0.2.1] leafNodes() crashes on first paint under better-sidebar 0.19 — "Cannot read properties of undefined (reading 'kind')"
## Symptom

In the DSH (DeepSeek Harness) web client, the Browser Skill sidebar tab **throws on first open**, every time. Host (dsh-better-sidebar) catches it in its error boundary and shows a "Retry" button; clicking Retry renders fine.

```
    at leafNodes  (@wxg-prc-cpg/browser-skill-dsh-plugin/lib/client.cjs)
    at observationTabOpen  (…)
    at evaluate  (…)
    at publish  (…)
```

## Root cause (file + line)

`lib/client.cjs` — `leafNodes()` (~L4462) and `observationTabOpen()` (~L4470):

```js
function* leafNodes(node) {
  if (node.kind === "leaf") {          // <-- throws when node is undefined
    yield node; return;
  }
  for (const child of node.children) yield* leafNodes(child);
}

function observationTabOpen(state) {
  if (state === void 0) return false;                      // only `state` is guarded
  for (const root of [state.splits, state.bottomSplits])   // either can be undefined on the first frames
    for (const leaf of leafNodes(root)) …
}
```

On the first render the host workbench layout state is not populated yet, so `state.splits` / `state.bottomSplits` are `undefined` → `node.kind` throws. That is why the first open always fails and the manual retry always succeeds.

## Suggested fix (one line)

```js
function* leafNodes(node) {
  if (node === void 0) return;
  if (node.kind === "leaf") { yield node; return; }
  for (const child of node.children ?? []) yield* leafNodes(child);
}
```

(or skip undefined roots inside `observationTabOpen`).

## Environment

* `@wxg-prc-cpg/browser-skill-dsh-plugin@0.2.1` (latest at the time of writing)
* `dsh-better-sidebar@0.19.1` (latest)
* DSH 0.1.5-rc.1 · Windows 11 · Node 24

## Related observation (same file, likely the same migration)

The tab is still opened with the legacy 0.18-era contract:

```js
service.openTab({ type: OBSERVATION_TAB_TYPE, path: OBSERVATION_TAB_PATH })
```

Under better-sidebar 0.19 this `path` appears to be resolved as a **file path**, producing
`cannot resolve target "…/browser-skill-observation": ENOENT` on the *auto-open* path
(manually clicking the tab renders fine). May be worth migrating to the 0.19 tab contract.

*Locally we patched `leafNodes` with an undefined-guard and verified the first-open crash is gone.*

L4XB:
Closes #232.

## What happened

Under `dsh-better-sidebar` 0.19 the workbench state reaches the plugin before either layout tree exists, so `observationTabOpen()` handed an `undefined` root to `leafNodes()`, which read `.kind` and threw `Cannot read properties of undefined (reading 'kind')`. The host's error boundary caught it, and the Browser Skill tab failed on every first open until the user pressed Retry (which always worked, because by then the layout was populated).

## Change

`leafNodes()` treats a missing node, and a split whose `children` are not populated yet, as having no leaves, so `observationTabOpen()` simply reports "not open" on those frames. `SidebarStateLike.splits` / `bottomSplits` are marked optional to document that the host can publish them absent; the only other reader, the test helper `withTabOpened`, handles that.

Not touched here: the related note in the issue about `service.openTab({ type, path })` resolving `path` as a file under the 0.19 tab contract. That is a separate contract migration and deserves its own change.

## Tests

New case in `tests/client/observation-sidebar.test.tsx`: an empty state, a state with only one of the two trees, a split without children next to a populated tree, and a lone childless split.

```
pnpm run build:css && pnpm run build:skill && npx vitest run tests/client/observation-sidebar.test.tsx   # 13 passed
npx tsc --noEmit   # clean
npx biome check <changed files>   # clean
```

### DIFF
--- a/packages/dsh-plugin-browserskill/src/client/observation-sidebar.tsx
+++ b/packages/dsh-plugin-browserskill/src/client/observation-sidebar.tsx
@@ -70,7 +70,8 @@ export interface SidebarStateLike {
    * optional here; earlier versions always provide it.
    */
   splits?: SidebarNodeLike;
-  bottomSplits: SidebarNodeLike;
+  /** Absent as well on the host's first frames, before either tree is populated. */
+  bottomSplits?: SidebarNodeLike;
   /** Whether the right panel is expanded (the merged drawer on narrow screens). */
   panelOpen?: boolean;
 }
@@ -164,12 +165,15 @@ export function ObservationSidebarTab({
   return <div className={css["sidebar-tab"]}>{body}</div>;
 }
 
-function* leafNodes(node: SidebarNodeLike): Generator<SidebarLeafLike> {
+function* leafNodes(node: SidebarNodeLike | undefined): Generator<SidebarLeafLike> {
+  // The workbench layout is not populated yet on the first paint, so a root
+  // (or a split's children) can be missing; an absent subtree has no leaves.
+  if (node === undefined) return;
   if (node.kind === "leaf") {
     yield node;
     return;
   }
-  for (const child of node.children) yield* leafNodes(child);
+  for (const child of node.children ?? []) yield* leafNodes(child);
 }
 
 /** Whether a tab of our type is already open in either sidebar workbench. */
--- a/packages/dsh-plugin-browserskill/tests/client/observation-sidebar.test.tsx
+++ b/packages/dsh-plugin-browserskill/tests/client/observation-sidebar.test.tsx
@@ -290,6 +290,23 @@ describe("observationTabOpen", () => {
     expect(() => observationTabOpen(noSplits)).not.toThrow();
     expect(observationTabOpen(noSplits)).toBe(false);
   });
+
+  it("tolerates a layout whose trees are not populated yet (first paint, #232)", () => {
+    // better-sidebar 0.19 publishes the state before either workbench tree
+    // exists; the first open used to throw "Cannot read properties of
+    // undefined (reading 'kind')" and land in the host's error boundary.
+    expect(observationTabOpen({})).toBe(false);
+    // Only one tree present: the tab is still found in it.
+    expect(observationTabOpen({ splits: leafWith(OBSERVATION_TAB_TYPE).splits })).toBe(true);
+    expect(observationTabOpen({ bottomSplits: leafWith(OBSERVATION_TAB_TYPE).splits })).toBe(true);
+    // A split whose children are not populated yet is treated as empty.
+    const halfBuilt: SidebarStateLike = {
+      splits: { kind: "split", id: "s1", dir: "row", sizes: [1] } as SidebarStateLike["splits"],
+      bottomSplits: leafWith(OBSERVATION_TAB_TYPE).splits,
+    };
+    expect(observationTabOpen(halfBuilt)).toBe(true);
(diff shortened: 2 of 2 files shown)


L4XB:
Rebased onto `main`. The conflict was with the DSH 0.1.5 change that made `splits` optional: both trees are optional now, with your comment kept for `splits` and one added for `bottomSplits`, and both tests are kept, the absent-right-column case from `main` and the first-paint case from this PR. `observation-sidebar.test.tsx` is at 14 passed.

sakkee37:
Thanks — this matches the first-paint crash I reproduced end-to-end on DSH Desktop 2.0.9 (core `0.1.5-rc.1`) with `dsh-better-sidebar` 0.19.x: it appears and disappears across reloads, and the host's error boundary is what surfaces it, so the message is prefixed with the host's name rather than this plugin's.

Closing my duplicate (#239) so the queue stays clean. If you want one more regression case, here is the one mine had that yours does not — a tree whose `children` array exists but carries a hole, asserting both that the walk does not throw **and** that it does not silently give up on tabs sitting next to the hole:

```ts
it("does not crash when the tree carries a missing child node", () => {
  const holed = {
    kind: "split",
    id: "s1",
    dir: "row",
    sizes: [1],
    children: [undefined, { kind: "leaf", id: "p3", active: null, tabs: [] }],
  } as unknown as SidebarNodeLike;
  expect(() => observationTabOpen({ bottomSplits: holed })).not.toThrow();
  expect(observationTabOpen({ bottomSplits: holed })).toBe(false);
});

it("still finds the tab when a sibling node is missing", () => {
  const holedWithTab = {
    kind: "split",
    id: "s2",
    dir: "row",
    sizes: [1],
    children: [
      undefined,
      {
        kind: "leaf",
        id: "p4",
        active: OBSERVATION_TAB_TYPE,
        tabs: [{ id: OBSERVATION_TAB_TYPE, type: OBSERVATION_TAB_TYPE, title: "Observation" }],
      },
    ],
  } as unknown as SidebarNodeLike;
  expect(observationTabOpen({ bottomSplits: holedWithTab })).toBe(true);
});
```

(`type SidebarNodeLike` may need adding to the existing import from `../../src/client/observation-sidebar`.) Take it or drop it as you prefer — either way, thanks for getting the guard in.

iuyo5678:
Thanks for the PR and the regression tests!
We’ve decided to proceed with #236, which integrates BrowserSkill directly with DSH’s newly introduced native right sidebar and provides a floating view as a fallback. This allows BrowserSkill to work with or without better-sidebar installed.
DSH and its plugin ecosystem are evolving rapidly. Tracking changes in both DSH and better-sidebar adds compatibility and maintenance overhead, so we’re choosing the more direct integration for now.
As DSH and the ecosystem mature, we may revisit deeper integration with popular plugins, including better-sidebar.
Since #236 removes the layout traversal that this PR protects, it will supersede this fix once merged. We appreciate your work investigating the issue and contributing a fix!

PR #239, opened 2026-09-13
fix(client): make the sidebar tree walk total (leafNodes)

sakkee37:
# fix(client): make the sidebar tree walk total so one malformed node cannot blank the sidebar

## Summary

`leafNodes()` in `src/client/observation-sidebar.tsx` dereferences `node.kind` and
iterates `node.children` without guarding either. The snapshot it walks comes from the
host at runtime, so a missing root **or a single hole in a stored split tree** throws
`TypeError: Cannot read properties of undefined (reading 'kind')`. Because
`observationTabOpen()` is called from a host store subscription, that throw is caught by
the host's render error boundary and takes the **whole right sidebar** down with a
"Retry" button.

This patch makes the walk total: `node === undefined` returns, and a missing
`children` list is treated as empty. It also adds a regression test for the holed-tree
case.

## Why the nested case still matters

`main` already guards the **root** (`observationTabOpen`: `if (root === undefined) continue`,
commit `fix(client): guard observationTabOpen against a missing right-column tree`,
2026-09-11 — thanks for that). That is enough for the path reported in #232, but the
walker itself is still partial, so the same crash is one schema/hole away. Guarding at
the walker covers both cases and keeps the invariant in one place.

## Reproduction / evidence

Environment where this reproduces (and where the fix was verified locally):

| item | value |
|---|---|
| DSH Desktop | 2.0.9 (macOS arm64), core `@deepseek-ai/* = 0.1.5-rc.1` |
| host | `dsh-better-sidebar` 0.19.0 **and** 0.19.1 |
| plugin | `@wxg-prc-cpg/browser-skill-dsh-plugin` **0.2.0** (npm `latest` 0.2.1 has the same unguarded walker) |

Observed behaviour: the crash is a **first-frame race** — it appears and disappears across
reloads (matching the "click Retry and it renders fine" reports in DSH-better-sidebar#620).
The two roots are `undefined` on the first frames, which is exactly the `node.kind`
dereference.

The error is misattributed by the host: `dsh-better-sidebar`'s `RenderBoundary` prefixes any
caught error with its own name, so this plugin's name never appears in the message — see
DSH-better-sidebar#620 and #639, where three reporters (and the host's own maintainer notes)
initially looked in the wrong repository. Independent stack captured in #620:

```
    at leafNodes           (@wxg-prc-cpg/browser-skill-dsh-plugin/client.js)
    at observationTabOpen
    at evaluate
    at publish
```

Local workaround verified on real hardware by @DCAMAR in #620:
`if (node === void 0) return;` + `node.children ?? []` — the same two guards as this PR.

## Verification

Two regression cases were added to `tests/client/observation-sidebar.test.tsx`:

- *"does not crash when the tree carries a missing child node"* — a `split` node whose
  `children` array contains an `undefined` entry must neither throw nor report an open tab.
- *"still finds the tab when a sibling node is missing"* — the guard must not degrade the
  walk into "give up on the whole tree": a tab sitting next to the hole is still found.

`leafNodes` is module-private, so there is no public API change.

I ran this repo's own gates on the branch before opening the PR:

| gate | command (repo's `lint` script) | result |
|---|---|---|
| format + lint | `biome check .` (v2.3.14, repo `biome.json`) | clean — *Checked 2 files. No fixes applied.* |
| types | `pnpm --filter @wxg-prc-cpg/browser-skill-dsh-plugin typecheck` | clean |
| tests | `pnpm --filter @wxg-prc-cpg/browser-skill-dsh-plugin test` | **237 passed** (14 files), incl. the two new cases |

Existing cases still pass: `finds the tab in either workbench tree`, `does not crash
when the right-column tree is absent (DSH 0.1.5+)`.

## Release request

`npm latest = 0.2.1` was published **2026-09-09**, two days *before* the root guard landed
on `main` (2026-09-11). So every user on the published version still hits the crash; the fix
is only reachable from `main`. A patch release (0.2.2) would let users pick it up without
patching `node_modules` by hand — that is the actual end-user fix for #232.


sakkee37:
Closing this as a duplicate of #235 — same walker change, opened two days earlier (2026-09-11), and it additionally documents the now-optional roots. No reason to keep two PRs for one fix in the queue.

For the record, this branch was verified against this repo's own gates before it was opened:

- `biome check` → clean
- `pnpm --filter @wxg-prc-cpg/browser-skill-dsh-plugin typecheck` → clean
- `pnpm --filter @wxg-prc-cpg/browser-skill-dsh-plugin test` → **237 passed** (14 files)

The one thing this branch had that #235 does not is a regression case for a split whose `children` array exists but carries a hole (`[undefined, leaf]`), including the positive assertion that a tab sitting next to the hole is still found. Rather than duplicate the fix, I've offered that case on #235.

PR #270, opened 2026-09-17
chore(deps): allow dsh 0.1.5-rc.1 alongside 0.1.0-rc.6

MrThanlon:
The plugin's peer and dev dependencies pinned the dsh packages to ^0.1.0-rc.6. Semver only matches prereleases of the same version tuple, so that range excludes the upcoming 0.1.5-rc.1 release and the plugin could not be loaded next to it. Widen the ranges to accept both lines and refresh the lockfile.


MrThanlon:
主要是 dsh 官方 desktop 严格检测 peer 版本匹配

MrThanlon:
0.1.6 换了插件管理方式, 不需要了

PR #281, opened 2026-09-18
fix(doctor): warn when the bundled agent skill is from an older bsk

### LINKED ISSUE #266: `bsk doctor` reports "agent skill up to date" even when the installed skill differs from upstream main - no version or source visibility
## Environment

- bsk CLI: 0.2.1 (`bsk-v0.2.1-x86_64-pc-windows-msvc.zip`), published 2026-09-09T06:34:56Z
- OS: Windows 11 x64
- Agent harness: WorkBuddy (`~/.workbuddy/skills/browser-skill/SKILL.md`)
- Extension connected, daemon protocol 1.1, the rest of `doctor` is fully green

## Summary

`bsk doctor` prints:

```
ok    agent skill up to date       up to date in 1 harness(es)
```

As far as I can tell, "up to date" here means only "the bundled skill is installed and CLI-managed". It does not mean "the installed skill matches upstream `main`". No version, commit, or provenance is shown anywhere, so an agent (or a human) reading `doctor` has no way to notice that the skill it is following is a week behind the repository.

## Evidence

| What | Size | Notes |
| --- | --- | --- |
| `skill/SKILL.md` at tag `cli-v0.2.1` | 9562 B | the version bundled into the 0.2.1 binary |
| Installed on disk | 9746 B | 9562 B + 184 CRLF (the file is 184 lines, all CRLF) -> byte-identical to the tag |
| `skill/SKILL.md` on `main` today | 15126 B | +5380 B vs installed |

The section structure is not a small edit, it is a rewrite:

```
installed (cli-v0.2.1)                    main (today)
---------------------------------------   -------------------------------
## Required lifecycle                     ## Before starting a session
## Work toward one observable goal        ## Task workflow
## Observe, act, observe                  ## Read and interact
## Respect the Agent Window boundary      ## Borrowing and browser settings
## Ask the human when needed              ## Human steps and recovery
## Command inventory                      ## Screenshots and Canvas
## File transfer                          ## Files and other tools
## Recover without wandering
```

`skill/SKILL.md` kept changing after the CLI release - the most recent commits touching it are `c8f9d63` (2026-09-16T11:47Z, "docs: reuse existing daemons and verify startup readiness"), `b433d37` (2026-09-16T11:30Z) and `3af3fe3` (2026-09-15T11:15Z).

## What I tried

I installed the upstream `main` file over the bundled one:

```
bsk install-skill --harness workbuddy --source <main SKILL.md> --force
```

On-disk size went 9746 -> 15126 B, and `doctor` still reported:

```
ok    agent skill up to date       synced 1 harness(es): workbuddy
```

So the check compares installed content against neither the bundled copy nor upstream. (I restored the bundled version afterwards and `doctor` went back to `up to date in 1 harness(es)`.)

## Why this matters for agents

The skill file *is* the operating manual the agent follows, and `doctor` is the only health signal it gets. When the two drift apart the failure mode is silent:

- The agent trusts `doctor` and assumes it holds the current instructions.
- An operator reading the repository sees different instructions than the agent is actually following, and reports stop reproducing.
- The concrete trap I hit: `main` documents `bsk screenshot --session <id> --full-page --out page.png` (around lines 199-218), but `bsk 0.2.1 screenshot --help` has no `--full-page` flag at all (only `--json --session --quiet --tab-id --ref -v --out`), so the call fails with `unexpected argument`. Nothing in `doctor` hints at that mismatch.

## Suggestions

Any one of these would have saved me the guesswork, roughly in the order I would value them:

1. **Stamp the skill with provenance.** Write a version/commit into the installed skill (frontmatter, or a sidecar file under `~/.bsk/`), and have `doctor` show it, e.g. `ok  agent skill  browser-skill 0.2.1 (bundled, cli-v0.2.1)`.
2. **Narrow the wording or add an explicit check mode.** `agent skill installed` or `skill present` would not imply "current". A `bsk install-skill --check` that reports bundled vs installed vs upstream without writing would be ideal for agents.
3. **Offer an update path.** `--source <path>` already works, it just requires the caller to fetch the file. Something like `bsk install-skill --from-upstream` (or having `bsk upgrade` also refresh the skill) would close the loop.
4. **Version-gate the docs.** Where `main` documents a flag that older CLIs lack (`--full-page` here), a "requires CLI >= x.y" note would avoid the `unexpected argument` dead end.

## Severity

Low - this is an observability gap rather than a broken flow, and everything else has been solid. Filing it mainly because `doctor` is the one command every agent runs, so its wording carries a lot of weight. Also related to #265, which I filed earlier today.

drakeo338:
Fixes #266.

The report is that `bsk doctor` prints `ok  agent skill up to date` while the
installed `SKILL.md` is actually stale. Reading `sync_installed_skills`, the
check isn't lying about what it measured — `up_to_date` means the on-disk
`SKILL.md` matched the copy **bundled in the running binary**. It's just that
"up to date" is a claim about the world, and what was verified is a claim about
one build. On an old CLI those two come apart, and the reporter's case is a good
illustration: 0.2.1 bundled a 9,562-byte skill, main had 15,126, and the
installed skill documented a `--full-page` flag that their own CLI didn't
support. Doctor said everything was fine.

That matters more than a normal wording nit because `doctor` is the first thing
an agent runs, and an agent that's told its instructions are current will go
ahead and follow them.

What I changed: when the check would previously report plain `ok`, it now
consults the update cache, and if a newer `bsk` is already known it returns a
warning that names both versions and points at `bsk update`.

A few things I deliberately didn't do.

I wanted to rename the check to something like "agent skill matches this bsk
build", which says what's actually compared. I didn't, because `CheckResult`
derives `Serialize` and `name` goes out in `bsk doctor --json`, so anything
matching on that string would break. The comments around `ok` vs `status` and
the `na` variant made it pretty clear this file treats its JSON shape as a
contract, so I left the name alone and changed the verdict instead.

I used `warn` rather than `fail`, because the skill genuinely is synced
correctly for the binary that's installed — nothing is broken, it's just older
than the user probably assumes. `warn` also keeps `ok: true`, so this can't flip
anyone's exit code.

I didn't add a network call. `print_update_hint_from_cache` already reads
`~/.bsk/update-check.json` read-only and offline, with the daemon as the only
writer, so I reused that path via a small `cached_newer_version` helper. It
keeps the existing freshness rule, which I think is the right default here: a
missing or stale cache reads as "don't know", and doctor stays quiet rather than
claiming the skill is stale when it can't tell.

And this doesn't compare anything against upstream `main` — it only tells you
your CLI is behind, which is the actual root cause of a stale bundled skill. The
version-stamping idea in the issue would go further, but that's a bigger change
and I'd rather not fold it into this one.

The one thing I'm unsure about is the wording of the detail line. It currently
reads `...; matches the skill bundled with bsk X, but bsk Y is available`, which
is accurate but long next to the other checks. Happy to shorten it if you'd
prefer.

Tests: three added in `doctor.rs` — that both previously-`ok` shapes (`updated`
and `up_to_date`) warn when a newer version is known, that `None` still reports
`ok` with no hint, and that a newer version doesn't downgrade a sync failure out
of `fail`. The existing `skill_check_from_report` callers pass `None`; I made
the newer version a parameter rather than reading the cache inside, so the tests
stay off the filesystem.

Verified with the commands CI runs: `cargo fmt --all -- --check` and
`cargo clippy --workspace --all-targets --locked -- -D warnings` are both
clean, and `cargo test --workspace --locked --no-fail-fast` goes from 708
passed / 1 failed to 711 passed / 1 failed — the three new tests, and the same
single failure before and after.

That one failure is `skill_install::sync::tests::sync_continues_on_partial_error`
and it's my environment, not your tree: it chmods a directory to `r-x` to force
a write error, and I'm running as root, which ignores the permission bits, so
the write succeeds and Codex ends up in `updated`. It fails the same way on a
clean checkout with my change stashed. Worth knowing if anyone else ever runs
the suite in a container as root.


iuyo5678:
Thanks for working on this. After reviewing current main, I’d prefer not to merge this change. Keeping the installed skill aligned with the running CLI’s bundle is expected, and normal CLI output already includes cached update notices. A newer CLI release also doesn’t necessarily mean the skill content has changed, so attaching that warning to the skill check adds limited value.

The remaining ambiguity is the “up to date” wording. A smaller follow-up clarifying that the skill matches the running CLI’s bundled version would address that more directly.

Thanks  for your PR  again!

PR #296, opened 2026-09-20
feat(extension): answer remote gateway UI requests outside the tool queue

lyingbug:
## Problem

A gateway that runs browser tasks for a user renders its own view of the running task: a small preview of the page the agent is working on, and a control that brings that tab to the front so the user can take over.

Neither can be built on the tool RPCs. `ToolQueues::dispatch` serializes work per session, so a `tool.screenshot` issued for a preview waits behind whatever the agent is doing; while `request_help` waits for the user, it waits for the whole human-help timeout. The result is that the preview goes dark exactly when the user most wants to look — during a long navigation, and while the agent is asking for help.

The DSH plugin's own PiP overlay works around this from the other side: `ObservationService.acquireForeground` cancels the pending thumbnail and kills an active `bsk screenshot` as soon as model-facing work starts, and only schedules a fresh frame after it ends.

## Change

Two optional requests are answered on the socket before `WSTransport` sees the frame, so they never enter the tool dispatcher:

```json
{"id":"ui-1","method":"ui.task_preview","params":{"session_id":"…"}}
{"id":"ui-2","method":"ui.task_focus","params":{"session_id":"…"}}
```

- `ui.task_preview` returns `image_base64`, `format`, `tab_id`, `title` and `captured_at`.
- `ui.task_focus` activates the task's own tab and raises its window.

Neither starts a session, and the channel is attached only to authenticated remote sockets — an extension connected to a local daemon answers `unknown_method` as before, so a gateway can feature-detect.

Constraints kept deliberately tight:

- Only a tab the task owns is captured or focused. Window membership alone never qualifies, so an unowned tab inside the Agent Window is skipped and another owned tab is used instead.
- Authorization, the document revision and the debugger attachment identity are re-checked after the capture and after the rescale; a frame that outlived its authorization is discarded rather than returned.
- The preview acquires the same explicit background-execution claim the tools use before reading a page that is not in the foreground.
- Captures are coalesced per task and at most one runs per tab. A local deadline cannot cancel a CDP command, so the fence is released only when Chrome answers; a poll that arrives meanwhile is refused instead of queueing another capture behind a stuck one. The caller's wait is bounded at 3s.
- UI frames keep the extension's control and help overlays visible: hiding and restoring them on every poll makes them flicker in the user's browser. Tool screenshots still suppress them for unobstructed page content.
- `session.stop` drains an in-flight capture before it detaches the debugger and closes tabs. No session state, turn-end detach or extra lifecycle callback is introduced.
- Frames are bounded to 640 pixels wide, encoded in the extension, because a viewport capture on a HiDPI display comes back several times larger than a preview needs.

`docs/remote-extension-connection.md` documents the contract and drops the sentence that said there is no preview or focus side protocol.

## Validation

- `pnpm --filter @browser-skill/extension test`: 1874 passed, 103 skipped. 12 of those are new, covering the channel (routing, invalid params, failure reporting, late replies on a closed socket) and the preview (overlay retention, coalescing, the HiDPI bound, the in-flight refusal and its recovery, stop draining a stuck capture, a failed stop restoring previews, unowned tabs, and discarding frames after authorization or attachment changes).
- `pnpm --filter @browser-skill/extension compile` passes.
- Not exercised against a live browser in this branch.

## Notes

The names are a proposal. If you would rather namespace these differently, or serve them from the daemon's remote server instead of the extension socket, say which shape you want and I will rework it.

This is the capability WeKnora has been carrying downstream; the version here is self-contained and free of anything specific to that deployment.


lyingbug:
Updated in 9cde489, with the review findings addressed as follows:

- **R1 — Preview versus tab return:** UI work is invalidated at the actual return/stop boundary and revalidated around acquisition. Late attach attempts remain fenced through rollback. Stale document pixels do not release the session-level claim shared with concurrent tools.
- **R2 — Window ownership:** Preview checks the actual tab window and current ownership before acquisition and before delivering the frame. A tab moved out of the Agent Window is rejected.
- **R3 — Focus versus teardown:** Focus uses the same lifecycle coordination as preview and rechecks the target between Chrome calls. Teardown waits up to one second for an already-issued focus mutation; if it remains pending, cleanup returns `cancelled` with reason `ui_busy` and can be retried without moving or closing the tab prematurely.
- **R4 — Bounded waiting:** The three-second UI deadline covers lookup, acquisition, capture and image processing. Stop does not wait indefinitely for image or screenshot promises. CDP cleanup bounds attachment/override waits and handles late completion without clearing a newer connection's claim. The screenshot fence remains tied to the underlying command and attachment; a response timeout does not cancel Chrome work.
- **R5 — Error reporting:** Socket responses preserve useful messages and distinguish unavailable targets, deadlines, busy captures, teardown cancellation and underlying failures through typed code/reason fields.

The protocol documentation now describes string IDs, remote-socket capability probing through `unknown_method`, shared claim lifetime and the limits of cancellation/cleanup.

Validation: **2122 extension tests passed, 108 skipped**; extension type checking and production build passed; repository lint passed, including **356 DSH tests**. The new regressions exercise the actual SessionManager, ChromiumCdp, tab-return, session-stop and multi-session disconnect paths with controlled Chrome timing, plus socket error envelopes. GitHub's frontend lint/typecheck/test/build job has also passed for this commit; other CI jobs are still running at posting time.

Live Chrome / remote-gateway end-to-end validation has not been run.

PR #297, opened 2026-09-20
feat(extension): attribute agent-opened tabs to the task that opened them

lyingbug:
## Background

Tab control has two shapes today. A local session may drive any tab inside its Agent Window; a remote session must own each tab explicitly, created through `tab_create` or borrowed with the user's confirmation (`authoriseAgentTab`, `tabs.ts`).

That leaves a gap for tabs a page opens by itself. When an agent clicks a `target="_blank"` link, or the page calls `window.open`, Chrome creates the tab and nothing attributes it to the task that caused it:

- A remote session sees it as `scope=user`. `tab_select`, `snapshot` and every page operation return `permission_denied`, and the only way forward is `tab_borrow`, which waits for the user to approve an overlay in that page — for an unattended task, the turn simply stalls. Links that open in a new tab are ordinary on search results, admin consoles, attachment links and OAuth flows, so this is not a rare corner.
- A local session is unaffected for a new tab in the same window, but a popup that opens its **own** window is denied with `agent_window_scope`, because it is no longer inside the Agent Window.

`docs/remote-extension-connection.md` records the gap: *"Automatic popup authorization is outside this version's scope."* This PR proposes closing it without weakening the authorization model.

## Change

The dispatcher wraps the tools whose page input can open a tab (`click`, `press_key`, `evaluate`, `fill`, `select`, and the three navigations) and watches Chrome's `webNavigation.onCreatedNavigationTarget` while the action runs.

A target is attributed only when **Chrome** reports it, **during** the action, from a tab the session already controls, and no other session owns it. `rel="noopener"` targets and nested popups qualify, because the attribution comes from Chrome's event rather than from `openerTabId`. A popup that opened its own window is moved into the Agent Window — which is the part that also fixes the local `agent_window_scope` denial.

Deliberately unchanged:

- Window membership and `openerTabId` never authorize a tab on their own. A tab the user drags into the Agent Window stays unauthorized for a remote session.
- A target Chrome reports without a source, or after the action ended, keeps the ordinary borrow flow with its user confirmation.
- One task cannot take a tab another task owns.
- A tab already inside the Agent Window still cannot be borrowed in place; that rule is untouched here.

The attribution window closes as soon as the action ends, with one exception: Chrome can deliver the navigation-target event just after acknowledging the input that caused it. The action therefore waits up to 100 ms — but only when a tab actually appeared while it ran, observed through `tabs.onCreated`. An action that opened nothing pays nothing, which keeps this off the hot path for ordinary clicks and typing.

## Rationale for the security shape

The agent already drives the source page and could navigate that same tab to the same URL, so attributing a target it just caused grants nothing it could not already reach. The alternative is not more safety: it asks the user to approve something the agent visibly just did, and that kind of reflexive prompt is what erodes the confirmations that matter — borrowing a tab the user already had open and signed into.

## Validation

- `pnpm --filter @browser-skill/extension test`: 1867 passed, 103 skipped, including 5 new cases — noopener and nested targets plus the popup-window move, a local session's popup window, refusing another task's tab and an unauthorized source, the event tail being paid only when a tab appeared, and listener cleanup when the action throws.
- `pnpm --filter @browser-skill/extension compile` passes.
- Not exercised against a live browser in this branch.

## Notes

The 100 ms tail is a heuristic for one Chrome ordering detail; if you know a signal that makes it unnecessary, I would rather use that. I am also happy to split the popup-window move into its own change if you prefer to take the local fix first.


lyingbug:
Updated in 7b74596. Main has also been merged and the dispatcher conflict resolved, preserving the upstream debug hooks and cancellation check around invocation.

- **R1 — Keyboard routing:** Replaced `tool.press_key` with the actual `tool.press` method. Dispatcher-level regressions verify that click and press both enter popup tracking.
- **R2 — Cancellation:** Popup tracking receives the same AbortSignal as the tool. Abort immediately removes target listeners and ends the optional event tail, even while invocation is still pending. Asynchronous continuations recheck cancellation before claiming or issuing a move. Claims legitimately established before cancellation remain recorded for normal cleanup; an already-issued Chrome move cannot be recalled.
- **R3 — Current source authority:** Both local and remote sources must still be in the Agent Window; remote sources also require explicit ownership. Source removal/detachment, session lifetime, existing target ownership and borrow reservations are checked across asynchronous validation. Pending candidates are separate from authorized tabs, so nested targets wait for their parent's validation and migration, and a rejected parent cannot authorize descendants.

Regressions cover cancellation during lookup and the 100 ms tail, moved local/remote sources, delayed parent migration, invalid parents, source closure/revocation, session termination, and targets claimed or reserved by another session while validation is pending.

The documentation now describes the policy as a browser-reported source relationship within a finite observation interval, rather than proof that a particular input caused the popup. The 100 ms tail remains bounded; late or unattributed targets retain the existing borrow flow.

Validation: **2102 extension tests passed, 108 skipped**; extension type checking and production build passed; repository lint passed, including **356 DSH tests**. GitHub's frontend lint/typecheck/test/build job has also passed for this commit; other CI jobs are still running at posting time.

Live-browser validation has not been run; the regressions use controlled Chrome events and API timing.

PR #299, opened 2026-09-20
fix(extension): treat overlay host as the click blocker

kevin9327:
## Problem
While the "Agent is controlling" overlay is visible, `bsk click` reports success but the page never receives the click. Uploads that click a file input fail the same way.

The overlay host is a full-viewport blocker (`:host([data-bsk-overlay-blocking]) { pointer-events: auto }`) mounted with a closed shadow root. Hit-testing only walked shadow children. Page JS cannot see a closed shadow, so the host itself was never treated as blocking, automation bypass never ran, and CDP mouse events hit the overlay instead of the page.

This matches https://github.com/Tencent/BrowserSkill/issues/277.

## Change
Inspect the overlay host's own box and `pointer-events` first. Shadow children remain a fallback for an open root.

## Testing
- `pnpm --filter @browser-skill/extension exec vitest run src/tools/__tests__/interaction.test.ts -t overlay`
- `pnpm --filter @browser-skill/extension test`

AI-assisted (Grok)


iuyo5678:
Once the CI errors are fixed, it can be merged; the other few PRs are similar.

Issue #219, opened 2026-09-10
passing the Chrome profile directly to the large language model, and the browser skill is ignoring the profile name on starting chrome


pleabargain:
I'm passing the Chrome profile directly to the large language model, and the browser skill is ignoring it. 
<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/718b1f9b-2fb5-4a0e-859a-07fadfc5a3f0" />

pleabargain:
I'm explicitly telling Browser Skill to use a specific profile, and it's ignoring that information! That is one of the key reasons I want to use the browser skill: it will use my existing profiles, but it is failing to do so. 
152.0.7977.84 (Official Build) (64-bit) (cohort: Control) 



error again

<img width="970" height="540" alt="Image" src="https://github.com/user-attachments/assets/6cbb24da-806d-49c9-b189-8a1c4fc919e2" />

pleabargain:
Sanitized — identifying references removed, technical detail preserved:

---

**Bug report: Browser Skill does not bind to a specific Chrome user-data profile**

Status: Open (workaround documented; product-side fix needed)
Severity: High — wrong-profile automation can publish to the wrong account context or block required work
Component: Browser Skill (bsk CLI + Chrome extension)
Affected workspace: [Redacted] — multi-profile setup requiring a specific Chrome profile
Date observed: 2026-09-10
Logged by: [Redacted]

Related local notes: [chrome-profile-browserskill.md](chrome-profile-browserskill.md) · preflight [../../scripts/assert-chrome-profile.ps1](../../scripts/assert-chrome-profile.ps1) · playbook [../../navigation-playbook.md](../../navigation-playbook.md)

### Summary

Browser Skill repeatedly connects to whichever Chromium instance has the extension in a Connected state, with no first-class way to select, pin, or verify a Chrome `--profile-directory` / user-data profile. In a multi-profile Windows Chrome setup, that silently selects the wrong profile (here: Other Profile instead of Required Profile), while agents and humans believe "browser is ready" because `bsk browsers` / the extension popup show Connected.

Compounding Chrome's process singleton behavior makes launching the correct profile from the CLI unreliable while another profile is already running. The combination produces a recurring failure mode: automation is blocked or aimed at the wrong identity until Chrome is forcibly restarted as the required profile and the extension is reconnected there.

### Environment

| Item | Value |
|---|---|
| OS | Windows 10/11 |
| Browser | Google Chrome 152.x |
[...]

PR #302, opened 2026-09-20
feat(extension): add human-readable browser instance names

imnemo:
Refs #219. Complements #298.

## Problem

`bsk browsers` already exposes a `label` and session selection already accepts `--browser <id-or-label>`, but the extension popup does not provide a way to set that label. With multiple connected profiles, users therefore have to map opaque instance IDs to browser profiles before every task.

## Changes

- Add a compact **Browser name** editor to the extension popup, with explicit save, Enter/Escape keyboard behavior, trimming, clearing, and a 48-character limit.
- Prevent labels that look like an 8-character instance ID, avoiding selector precedence surprises.
- Reconnect after a persisted label change so the daemon registry sees the new name immediately. The popup explains that saving ends active tasks in that browser.
- Preserve instance IDs as the canonical fallback and keep duplicate-label handling fail-closed in the daemon.
- Add English, Simplified Chinese, and Korean copy, plus profile-selection documentation and CLI examples.

No protocol changes or new browser permissions are required. Automatic Chrome profile discovery remains out of scope; users still choose the mapping in the profile where the extension is installed.

## Validation

- Extension test suite: **1,882 passed / 103 skipped**
- Focused popup and connection lifecycle tests: **58 passed**
- Localization tests: **52 passed**
- Extension TypeScript compilation: passed
- Production extension build: passed
- Biome and Stylelint: passed
- Manual popup review at the shipped 320–340 px width: default, valid-name, and instance-ID-shaped validation states checked

`pnpm lint` currently reaches the DSH plugin suite and then fails on the existing prompt-size assertion (`8,133 < 8,000`). The same failure is present on upstream `main` at `0225683` and in its [CI run](https://github.com/Tencent/BrowserSkill/actions/runs/35511970480); this change does not touch the DSH plugin or generated skill content.


iuyo5678:
Thank you first of all for submitting the PR. I have a small suggestion for your consideration.
Currently, in order for the new name to take effect immediately, the code logic reconnects the extension, terminates all active tasks in the browser, and triggers session cleanup. For example, if the Agent is filling out a form, uploading, or collecting content, and the user simply changes the name, the task will also be interrupted.

Would it be possible to adjust this so that: when there are no tasks, it takes effect immediately; when there are tasks, it takes effect on the next connection, or, more simply, saving could be temporarily disabled. 
Wouldn't this better match the general user expectation for “changing the display name”.

imnemo:
Thanks for the suggestion — agreed that renaming should never interrupt an active browser task. Updated in `69dfd06`:

- the popup now receives the live session count and disables both the name field and Save button while any session is active;
- the background rechecks the session count before persisting, which closes the race where a task starts from a stale popup;
- session start/stop changes refresh the popup immediately, so editing becomes available as soon as the browser is idle;
- when idle, saving still reconnects briefly so the daemon sees the new name immediately;
- the busy-state copy and documentation were updated in English, Simplified Chinese, and Korean.

Validation after rebasing onto current `main`: 1,890 extension tests passed (103 skipped), 65 focused tests passed, 52 localization tests passed, repository lint passed including 351 DSH tests, and extension compile/build passed.

